PR #1, opened 2021-01-25
Copy tokens from Kubernetes to the Secret Manager.

lgruen:
There's still a lot of functionality missing overall, but I prefer sending a number of smaller PRs.


pdiakumis (approved):
Thank you - I don't fully understand the context but we can discuss next week.

lgruen:
> Thank you - I don't fully understand the context but we can discuss next week.

Sorry, I forgot to include the top-level `README.md` file! I'll send that as another PR.

PR #2, opened 2021-01-26
Add top-level readme file.

lgruen:


pdiakumis (approved):
Awesome, that makes sense, thank you!

PR #5, opened 2021-01-31
Update lint configs.

lgruen:


.pre-commit-config.yaml
```diff
@@ -5,17 +5,23 @@ repos:
...
   rev: 20.8b1
   hooks:
   - id: black
 
-- repo: https://github.com/pycqa/pylint
-  rev: '2.6'
-  hooks:
-  - id: pylint
-    additional_dependencies: [pylint-quotes]
+# Currently disabled, as dependencies are managed through conda.
```
vladsaveliev, line 19: But conda only takes care of installing pylint, but not of running as a pre-commit hook. Why removing it here?

.github/workflows/main.yaml
```diff
@@ -4,16 +4,24 @@ on: [push, pull_request]
...
-        run: pip install pylint pylint-quotes flake8 flake8-bugbear black
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          environment-file: environment.yml
+          activate-environment: analysis-runner
 
       - name: pylint
-        run: pylint **/*.py
+        run: |
+            pip install pylint-quotes  # TODO: remove this once available in conda
```
vladsaveliev, line 23: I'll use flake8-quotes instead as it's available on conda-forge

.pre-commit-config.yaml
lgruen: Sorry, that comment was a bit obscure :). pylint also checks the Python imports. In this case it complains e.g. about `kubernetes` not being known, because that's not listed as one of the pylint dependencies. It would be awesome if we could use the same conda environment file that's already used for the GitHub workflow to also perform the pre-commit checks.

.github/workflows/main.yaml
lgruen: Great!

.pre-commit-config.yaml
lgruen: Updated all of this with your new configs now :).

vladsaveliev (approved):

PR #9, opened 2021-02-04
Remove /deploy-config from docker-file

illusional:
- Remove from dockerfile
- Delete config from `/driver/`


lgruen (approved):
Thank you!

PR #11, opened 2021-02-08
Move dataset stack deployment code to the analysis runner repository.

lgruen:


stack/Pulumi.fewgenomes.yaml
```diff
@@ -0,0 +1,8 @@
+encryptionsalt: v1:SPQXlVggjxw=:v1:X1sTpViNuyK+Wcom:3GANI8gZOKtk0gG7BklsyHeNU5uVLw==
+config:
+  datasets:customer_id: C010ys3gt
+  datasets:enable_release: "false"
+  datasets:hail_service_account: fewgenomes-911@hail-295901.iam.gserviceaccount.com
```
illusional, line 5: Are these okay to be public facing? I know they're just service accounts.

stack/__main__.py
```diff
@@ -0,0 +1,391 @@
+"""
+Pulumi stack to set up buckets and permission groups
+"""
+
+import pulumi
+import pulumi_gcp as gcp
+
+DOMAIN = 'populationgenomics.org.au'
+CUSTOMER_ID = 'C010ys3gt'
```
illusional, line 9: Another is this okay to be public facing.

stack/__main__.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+def group_mail(kind: str) -> str:
+    """Returns the email address of a permissions group."""
+    return f'{dataset}-{kind}@{DOMAIN}'
+
+
+def create_group(mail: str) -> gcp.cloudidentity.Group:
+    """Returns a new Cloud Identity group for the given email address."""
+    name = mail.split('@')[0]
```
illusional, line 118: Should we process the name further, by maybe making it lowercase with `.lower()`?

stack/Pulumi.fewgenomes.yaml
lgruen: Thanks for checking! Yes, I don't think somebody else could use this information without having additional permissions.

stack/__main__.py
lgruen: Ditto.

stack/__main__.py
lgruen: The name here is just a Pulumi identifier that's displayed in the plan. So there's no need to do a lot of processing, it just was a bit long when having the domain attached to it.

illusional (approved):
Comments address my issue, and seems pretty good!

Issue #8, opened 2021-02-02
Add CLI tool


lgruen:
We should have a simple CLI tool that gathers some of the parameters for the request that's sent to the server automatically: e.g. GitHub repository, commit hash. It should also map a human-readable project name to the correct HTTPS endpoint.

illusional:
Hey Leo, just putting together a rough plan for actioning this issue, does this match up with what you were expecting?

A CLI called like:

```
analysis-runner run \
     --output-dir gs://some-path/ \
    [--commit-hash <current-hash>] \
    [--repo <current-repo>]  \
    --script <maybe default: main.py with no args>
```

Process:

1. Fill in the mising info (repo, commit hash, etc)
2. Get the gcloud auth token $(gcloud auth print-identity-token)
3. Form a POST request with params:
    * output
    * repo
    * commit
    * script
    * description
4. Collect and print response

illusional:
Further to this discussion, we need to implement some mechanism for deciding the server. There should be a `--stack` parameter used for determining the server, this is simply looked up in a JSON file hosted on GH. If this tool is used more frequently, we could setup a cloud function to return the correct server.

illusional:
Closed by #12.

PR #12, opened 2021-02-09
Add CLI for analysis runner

illusional:
Addresses #8.

This is just the code for the CLI, there's currently no `setup.py` so that it's installable. More so you could directly run this script:

```
cli/cli.py [...params]
```


cli/cli.py
```diff
@@ -0,0 +1,155 @@
+"""
+CLI for interacting with the REMOTE analysis-runner
```
lgruen, line 2: Can we move this to a README.md instead? I find that a bit more discoverable than in the source code.

cli/cli.py
```diff
@@ -0,0 +1,155 @@
...
+BRANCH = 'add-cli'  # 'master'
+DEFAULT_STACK_LOOKUP = (
+    f'https://raw.githubusercontent.com/'
+    f'populationgenomics/analysis-runner/{BRANCH}/cli/stacklookup.json'
+)
+
+
+@click.command()
+@click.option(
+    '--stack',
```
lgruen, line 44: "Dataset" might be easier to understand for analysts, as that's the term we use in the storage policies. Using a Pulumi stack is just the concrete technical implementation.

I'm not super happy with "dataset" either, but "project" is already terribly overloaded. Let me know if you have a better idea :).

cli/cli.py
```diff
@@ -0,0 +1,155 @@
...
+    '--repository',
+    help='The URI of the repository to run, must be approved by the appropriate server.'
+    ' Default behaviour is to find the repository of the current working'
+    ' directory with `git remote get-url origin`',
+)
+@click.option(
+    '--commit-ref',
+    '--hash',
+    help='The hash of repository to run, default behaviour is to '
+    'find the latest hash of the current repository',
```
lgruen, line 63:
```suggestion
    help='The hash of commit to run, default behaviour is to '
    'use the current commit',
```

I think it would be confusing to use HEAD here instead of whatever the user has checked out currently.

cli/cli.py
```diff
@@ -0,0 +1,155 @@
...
+)
+@click.option(
+    '--description',
+    help='Description of job, otherwise defaults to: "$USER FROM LOCAL: $REPO@$COMMIT"',
```
lgruen, line 67: We already log user, repo, etc., so that doesn't need to be part of the description. This should be something that adds extra context, like a git commit message. I think it might make sense to make this mandatory, but I don't have a strong opinion.

cli/cli.py
```diff
@@ -0,0 +1,155 @@
...
+)
+@click.option('--script', multiple=True, default=['main.py'])
```
lgruen, line 69: Should we really allow multiple here?

illusional: The `script` is just an array of <script> and params. It's handled as an array, I've actually modified the command line to be:

```python
@click.option('script', nargs=-1, default=['main.py'])
```

So you can use like:

```sh
cli.py [...params] main.py <script-arg1> ... <script-argn>
```
Which becomes:
```
script = ["main.py", "<script-arg1>", ..., "script-argn"]
```
```

File: cli/cli.py (hunk 4 above)
Comment (illusional) [reply cli/cli.py:67]: I've made `description` a required field and removed the default.

File: cli/cli.py (hunk 2 above)
Comment (illusional) [reply cli/cli.py:44]: I agree staying away from _project_, I think dataset is a reasonable middle-ground, hence I've renamed it to `dataset`.

File: cli/cli.py (hunk 1 above)
Comment (illusional) [reply cli/cli.py:2]: Yep, I've created a README

File: README.md [hunk 6]
```diff
@@ -19,6 +19,37 @@ used to run the batch on your behalf. There's a dedicated Batch service
 account for each project (e.g. "tob-wgs"), which helps with bucket permission
 management and billing budgets.
 
+## Access through CLI
+
+If you have the `analysis-runner` CLI installed, you can request the analysis-runner
```
lgruen, line 24: Can we add a link here to the installation instructions? (E.g. in the getting started guide.)

README.md
```diff
@@ -19,6 +19,37 @@ used to run the batch on your behalf. There's a dedicated Batch service
...
+starts jobs from a GitHub repository + commit.
```
lgruen, line 25:
```suggestion
to start pipelines based on a GitHub repository, commit, and command to run.
```

README.md
```diff
@@ -19,6 +19,37 @@ used to run the batch on your behalf. There's a dedicated Batch service
...
+
+If you're in the directory of the project you want to run, you can omit
+the `--commit` and `--repository` parameters, which will use your current REMOTE
+and commit HEAD.
+
+For example:
+
+```bash
+cpg-analysisrunner \
```
lgruen, line 34: Can we call this analysis-runner, for consistency with the repository name?

README.md
```diff
@@ -19,6 +19,37 @@ used to run the batch on your behalf. There's a dedicated Batch service
...
+    --dataset <data-set> \
```
lgruen, line 35:
```suggestion
    --dataset <dataset> \
```

README.md
```diff
@@ -19,6 +19,37 @@ used to run the batch on your behalf. There's a dedicated Batch service
...
+    --commit <commit-sha> \
+    --dataset <data-set> \
+    --description <description> \
+    --output-dir gs://<bucket-path> \
+    script_to_run.py with arguments
+```
+
 ## Usage
 
 **TODO(@lgruen):** Add instructions
```
lgruen, line 55: Remove this TODO?

cli/README.md
```diff
@@ -0,0 +1,38 @@
+# Analysis Runner CLI
+
+CLI for interacting with the REMOTE analysis-runner (externally hosted `server/main.py` script).
+
+Motivation: [Source](https://github.com/populationgenomics/analysis-runner/issues/8)
+
+> A simple CLI tool that gathers some of the parameters for the request that's sent to the
+> server automatically: e.g. GitHub repository, commit hash. It should also map a
+> human-readable project name to the correct HTTPS endpoint.
+
```
lgruen, line 10: Probably no need to cite this.

cli/README.md
```diff
@@ -0,0 +1,38 @@
...
+Called with:
+
+```bash
+analysis-runner run \
+    --dataset <ds-id> \
```
lgruen, line 15:
```suggestion
    --dataset <dataset> \
```

cli/README.md
```diff
@@ -0,0 +1,38 @@
...
+    script.py to run with args
+```
+
+## CLI Overview
+
+Process:
+
+1. Fill in the mising info (repository, commit hash)
+2. Get the gcloud auth token
+    * `BASH: $(gcloud auth print-identity-token)`
```
lgruen, line 28:
```suggestion
    * `gcloud auth print-identity-token`
```

cli/cli.py
```diff
@@ -0,0 +1,206 @@
...
+import google.auth.transport.requests
+
+logging.basicConfig(level='INFO')
+
+BRANCH = 'main'
+DEFAULT_SERVER_LOOKUP = (
+    f'https://raw.githubusercontent.com/'
+    f'populationgenomics/analysis-runner/{BRANCH}/cli/servermap.json'
+)
+# Make this None if you want to support all organisations
```
lgruen, line 23: Can we use the American spelling? (GitHub and Google use organization.)

cli/cli.py
```diff
@@ -0,0 +1,206 @@
...
+            "when specifying the '--repository'"
+        )
+
+    _repository = repository or _get_default_remote()
+    _commit_ref = commit or _get_commit_ref_of_current_repository(
+        used_custom_repository=repository is not None
+    )
+    _url = _get_url_from_dataset(dataset)
+    _token = _get_google_auth_token()
+    _script = script or ['main.py']
```
lgruen, line 80: We should grab the current working directory here. I.e. if my repository root is at `/my/repo` and I'm currently in `/my/repo/analysis/project` and pass `main.py` as the script, the command should become `analysis/project/main.py`.

cli/cli.py
```diff
@@ -0,0 +1,206 @@
...
+
+    logging.info('Submitting %s@%s for dataset "%s"', _repository, _commit_ref, dataset)
```
lgruen, line 82: How do you feel about using f-strings here (and below). To me they're a bit more readable.

cli/cli.py
```diff
@@ -0,0 +1,206 @@
...
+            response.status_code,
+            str(e),
+            response.text,
+        )
+
+
+def _get_google_auth_token() -> str:
+    # https://stackoverflow.com/a/55804230
+    # command = ['gcloud', 'auth', 'print-identity-token']
+
```
lgruen, line 110:
```suggestion
```

cli/cli.py
```diff
@@ -0,0 +1,206 @@
...
+    raise Exception(f"Couldn't get URL for '{dataset}', expected one of {d.keys()}")
+
+
+if __name__ == '__main__':
+    # Disable pylint because click decorates the function in a specific way
+    # pylint: disable=no-value-for-parameter
+    main()
+
+# if __name__ == '__main__':
```
lgruen, line 202: Remove this?

cli/cli.py
```diff
@@ -0,0 +1,206 @@
...
+    '--repository',
+    '--repo',
+    help='The URI of the repository to run, must be approved by the appropriate server.'
+    ' Default behaviour is to find the repository of the current working'
+    ' directory with `git remote get-url origin`',
+)
+@click.option(
+    '--commit',
+    '--hash',
+    help='The hash of commit to run, default behaviour is to ' 'use the current commit',
```
lgruen, line 54:
```suggestion
    help='The hash of commit to run, default behavior is to use the current commit',
```

cli/cli.py
```diff
@@ -0,0 +1,206 @@
...
+@click.option(
+    '--repository',
+    '--repo',
+    help='The URI of the repository to run, must be approved by the appropriate server.'
+    ' Default behaviour is to find the repository of the current working'
+    ' directory with `git remote get-url origin`',
+)
+@click.option(
+    '--commit',
+    '--hash',
```
lgruen, line 53:
```suggestion
```

cli/cli.py
```diff
@@ -0,0 +1,206 @@
...
+)
+@click.option(
+    '--repository',
+    '--repo',
+    help='The URI of the repository to run, must be approved by the appropriate server.'
+    ' Default behaviour is to find the repository of the current working'
+    ' directory with `git remote get-url origin`',
+)
+@click.option(
+    '--commit',
```
lgruen, line 52: Might be worth mentioning that this also supports tags. However, `HEAD` is not allowed (and we should probably add more checks to e.g. disallow `HEAD^` and similar).

cli/servermap.json
```diff
@@ -0,0 +1,4 @@
+{
+  "tob-wgs": "https://analysis-runner-server-9eddd6b-aceka37pgq-ts.a.run.app",
+  "local": "http://localhost:3000"
```
lgruen, line 3:
```suggestion
  "fewgenomes": "https://analysis-runner-server-bc3254b-i5utivowgq-ts.a.run.app"
```

cli/requirements.txt
```diff
@@ -0,0 +1,2 @@
+click
```
lgruen, line 1: Should we pin versions here?

server/main.py
```diff
@@ -129,7 +129,9 @@ async def index(request):
         job.command(f'git merge-base --is-ancestor {_shell_escape(commit)} HEAD')
         job.command(f'git checkout {_shell_escape(commit)}')
         # Make sure the file is in the repository.
-        script_file = script.partition(' ')[0]
+        if not isinstance(script, list):
```
lgruen, line 132: Should we really support different input types here? This will later be converted into a single string anyhow, so maybe we should just pass a string from the CLI too? Supporting a list here complicates the logic below where we split by spaces for the escaping.

lgruen (changes requested):

server/main.py
illusional: The current implementation won't support parameters with spaces in it, eg: "this is a single param with spaces". We get an array straight from the argument parser, and it's more convenient for the server to have an array anyway based on how arguments are quoted.

lgruen: Won't you have the same problem though once you assemble the command for Hail Batch to run? I.e. `Job.command` doesn't support a list either.

cli/cli.py
illusional: 👍  I much prefer f-strings but they were blocked by pylint, I'll disable `logging-fstring-interpolation` pylint checks in `analysis-runner` and revert these back.

cli/cli.py
illusional: I'd prefer to keep this here, I think provides clear context for the motivation of the function.

server/main.py
illusional: The original code splits by the space character, and then shell quotes each element. I've just moved this up higher, and added support for if you supply the array directly:

Original:
https://github.com/populationgenomics/analysis-runner/blob/4ea28daeab2c5480749024c1e1cae5f60f225ab9/server/main.py#L158-L159

README.md
illusional: Yep, I've added:

> Please see the [Team Docs](https://github.com/populationgenomics/team-docs/) for
installation instructions.

And when it's deployed (and I can double check the actual install instructions), I'll add them to the team docs. I've avoided linking to a specific page in case links move.

README.md
illusional: Renamed to `analysis-runner`.

cli/cli.py
illusional: Murica'd it up in: 773fcf724bbbc3f29637ab9326d9a6e4675a2462

server/main.py
lgruen: Got it, thanks!

cli/cli.py
illusional: Sweet! The script does something like this now:

```python
if _script[0]:
    # Make the first argument (the script name) relative
    # to the git root and current directory
    _script[0] = _get_relative_script_path_from_git_root(_script[0])
```

cli/requirements.txt
illusional: Do we have a general style guide on pinning repositories? I prefer not to, because I've had bad experience with contextual version conflicts being a pain, I can also set a large range? Or down to a specific version if that's what's desired.

illusional:
Pinging @vladsaveliev, there are issues with the yaml linter failing because of the conda files:
```
Check Yaml...............................................................Failed
- hook id: check-yaml
- exit code: 1

while scanning for the next token
found character that cannot start any token
  in "conda/analysis-runner/meta.yaml", line 1, column 2
```

Any suggestions for this?

cli/requirements.txt
lgruen: We don't have a guide yet (apart from analyses, where we recommend pinning versions for reproducibility). The disadvantage of not pinning versions here is that one day this might stop compiling for an unrelated commit, because of a breaking API change in one of the dependencies. As this is a standalone tool, would there be any drawback when pinning to whatever is latest right now?

README.md
```diff
@@ -19,9 +19,38 @@ used to run the batch on your behalf. There's a dedicated Batch service
...
+    script_to_run.py with arguments
+```
+
+If you provide a `--repository`, you MUST supply a `--commit <SHA>`, eg:
+
+```bash
+analysis-runner \
+    --repository my-approved-repo \
+    --commit <commit-sha> \
+    --dataset <data-set> \
```
lgruen, line 49:
```suggestion
    --dataset <dataset> \
```

cli/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+If you provide a `--repository`, you MUST supply
+a `--commit <SHA / or tag>`, eg:
+
+```bash
+analysis-runner \
+    --repository my-approved-repo \
+    --commit <commit-sha> \
+    --dataset <data-set> \
```
lgruen, line 24:
```suggestion
    --dataset <dataset> \
```

cli/cli.py
```diff
@@ -0,0 +1,228 @@
...
+import subprocess
+
+import click
+import requests
+import google.auth
+import google.auth.transport.requests
+
+logging.basicConfig(level='INFO')
+
+BRANCH = 'add-cli'
```
lgruen, line 19:
```suggestion
BRANCH = 'main'
```

Or hardcode it directly in the lookup below.

lgruen (approved):

cli/requirements.txt
illusional: Pinned to:

```
google-auth ==1.25.0
click ==7.1.2
```

server/main.py
```diff
@@ -61,6 +61,8 @@ def _read_secret(name: str) -> str:
     return response.payload.data.decode('UTF-8')
 
 
+# The extra number of params is from @routes.post, so we'll ignore the error
```
vladsaveliev, line 64: Could probably edit or remove this message since it was about statements?

server/main.py
```diff
@@ -129,7 +131,9 @@ async def index(request):
         job.command(f'git merge-base --is-ancestor {_shell_escape(commit)} HEAD')
         job.command(f'git checkout {_shell_escape(commit)}')
         # Make sure the file is in the repository.
-        script_file = script.partition(' ')[0]
+        if not isinstance(script, list):
+            script = str(script).split(' ')
+        script_file = script[0]
```
vladsaveliev, line 136: It's not obvious from the code that the `script` parameter can contain command-separated arguments. Maybe could move the parsing to 109, and make it like this?

```
script = params['script']
if not script:
    raise web.HTTPBadRequest(reason='Invalid script parameter')
if isinstance(script, str):
    script = script.split(' ')
script_file, *script_args = script
```

Plus, I'd split on any whitespace (`script.split(' ')` -> `script.split()`) to allow for multiple spaces.

vladsaveliev (approved):

PR #14, opened 2021-02-11
Conda fixes & add bump2version

vladsaveliev:
- Added bump2version into the dev environment to manage the version of the CLI package. Documented the usage in the README.
- Enable the `--version` flag for the CLI
- A few fixes to the conda package (`google-auth` was only available as 1.24.0; the test was wrong; `host` deps were missing)
- Slightly updated the docs to reflect the now-available conda package. I copied the usage from `--help` just so the following sentence in the docs follows something ("here is the usage, but you can omit the `--commit` and `--repository` parameters..."), and I think it's a good practice anyway.
- Few fixes in docs (e.g. the dev environment yaml is called `environment-dev.yml`).


setup.py
```diff
@@ -9,7 +12,7 @@
 
 setuptools.setup(
     name=PKG,
-    version='0.1.2',
+    version='1.0.0',
```
illusional, line 15:
```suggestion
    # This tag is automatically updated by bump2version
    version='1.0.0',
```

vladsaveliev: Yes, absolutely should have this warning here, thanks for the suggestion! I'll commit it, but feel free to parse `_verison.py` here instead if it works better for you.

illusional (approved):

Issue #7, opened 2021-02-02
Send metadata to Airtable


lgruen:
Using our automation account, send the run metadata to an Airtable base. This way, we can link runs to data outputs (whereas data outputs are already linked to runs through the `metadata.json` file created in the output directory).

lgruen:
Might be a good idea to use a Pub/Sub topic for this.

PR #15, opened 2021-02-11
Send metadata to Pub/Sub

lgruen:
That's the first half of #7.


illusional (approved):
Awesome, love when things can talk to each other!

PR #16, opened 2021-02-11
CLI: Only make script path relative if repository is None

illusional:
Fix error when providing repository

### DIFF
--- a/cli/cli.py
+++ b/cli/cli.py
@@ -78,15 +78,15 @@ def main(dataset, output_dir, script, description, commit=None, repository=None)
     _commit_ref = commit
     _script = list(script)
 
-    if repository is None:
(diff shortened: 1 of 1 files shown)


cli/cli.py
```diff
@@ -160,7 +160,7 @@ def _get_relative_script_path_from_git_root(script_name: str) -> str:
 
 def _get_git_default_remote() -> str:
     command = ['git', 'remote', 'get-url', 'origin']
-    full_remote = _get_output_of_command(command, 'get GIT repository')
+    full_remote = _get_output_of_command(command, 'get default GIT repository')
```
lgruen, line 163: Nit here and below: I think Git is usually spelled not in all caps.

lgruen (approved):

PR #17, opened 2021-02-11
Add Pub/Sub IAM permissions

lgruen:


stack/__main__.py
```diff
@@ -321,7 +322,7 @@ def add_bucket_permissions(
                     ],
                     image=(
                         f'australia-southeast1-docker.pkg.dev/analysis-runner/'
-                        f'images/server:78d20393125b'
+                        f'images/server:650b5c0505c6'
```
illusional, line 325: This value changes a bit, is it worth making it a constant at the top so you don't have to hunt for it when it updates?

illusional (approved):

stack/__main__.py
lgruen: Good idea. Done in https://github.com/populationgenomics/analysis-runner/pull/20/files.

PR #21, opened 2021-02-12
Bump version: 1.0.1 → 1.0.2

illusional:
I think maybe the v1.0.1 release wasn't commited. Because it was released here: https://github.com/populationgenomics/analysis-runner/releases/tag/v1.0.1

So I've bumped to v1.0.2


illusional:
If this is approved, I'll just push the tag that bump2version created

vladsaveliev:
I pushed the tag (`git push --tags`), but forgot to make a standard push commit (`git push`, or in our case should have created a branch and a PR). Good catch!

The package was uploaded though: https://anaconda.org/cpg/analysis-runner

vladsaveliev (approved):

PR #30, opened 2021-02-23
Check permission groups explicitly, allowing for extended access

lgruen:
It also simplifies the deployment by only running a single analysis-runner server instance.


illusional (approved):
Awesome, I think this is a great change :)

PR #33, opened 2021-02-24
Clean up containers

vladsaveliev:
Clean up conda package source tarballs as the last step of the driver and server images build. Reduces the size of the drive image by 1.9G.


lgruen (approved):
Woah! Nice!!

Issue #37, opened 2021-02-27
Continuous deployment


lgruen:
Updating version tags manually is tedious and error-prone. We should try to get to the following state:

- [x] When a new Hail version is built on conda, the driver image should get built with the new version (https://github.com/populationgenomics/analysis-runner/blob/main/driver/Dockerfile#L16).
- [x] When a new driver image gets built, the server image should get built with the new version (https://github.com/populationgenomics/analysis-runner/blob/main/server/Dockerfile#L2 and https://github.com/populationgenomics/analysis-runner/blob/main/server/Dockerfile#L2).
- [x] When a new server image gets built, the server should get redeployed.

For the first point we'd need to figure out how to trigger a workflow across repositories.

For all of these, we should still update versions explicitly instead of referring to "latest". These would need to be automated commits to version files.

lgruen:
We can't do automated commits due to our branch protection rules. Instead, https://github.com/populationgenomics/analysis-runner/pull/38 uses an environment variable to set the version. This should be okay, since the version is also logged to Airtable for each run.

PR #38, opened 2021-03-04
Add a workflow for continuous server deployment

lgruen:
See #37 for context.

This workflow should be triggered whenever a new Hail package gets released to conda (see https://github.com/populationgenomics/hail/pull/72).

It builds a new driver image, server image, and deploys to Cloud Run.

Example run: https://github.com/populationgenomics/analysis-runner/runs/2028560588


illusional (approved):
Awesome, so cool to see this automation being spun up.

PR #42, opened 2021-03-05
Add end-to-end test

### LINKED ISSUE #39: Add end-to-end tests
When redeploying the analysis server, we should first test whether submitting a batch to a test instance succeeds. Only then should the production instance be updated.

lgruen:
Fixes #39.

### DIFF
--- a/test/main.py
+++ b/test/main.py
@@ -1,19 +1,5 @@
 """A very simple batch that tests basic functionality."""
 
-import os
 import hail as hl
-import hailtop.batch as hb
 
 hl.init()
-
-backend = hb.ServiceBackend(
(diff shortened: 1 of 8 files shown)


test/main.py
```diff
@@ -0,0 +1,5 @@
+"""A very simple batch that tests basic functionality."""
+
+import hail as hl
+
+hl.init()
```
illusional, line 5: Does this do enough to test if it works, or would we want to actually print something, like the most minor of "Hello, World" of calculations?

illusional (approved):

test/main.py
lgruen: At that point we're kind of testing the Hail backend -- which seems somewhat out of scope for the analysis runner.

A bigger problem is that the deployment of Hail is generally running _in parallel_. Ideally, we'd wait until the `prod deploy` finished and then run this test, but that's tricky.

PR #43, opened 2021-03-08
Use shlex.quote instead of _shell_escape

lgruen:


illusional (approved):
I don't mind the use of shlex.quote here, but I had problems with it in my larger batch workflows because of its use of the single quote. 

I'll talk to you more about this in our one-on-one but this is good :)

PR #45, opened 2021-03-14
Attempt to fix gunicorn setup

lgruen:
Cloud Run couldn't find the gunicorn executable before.


vladsaveliev (approved):
That makes sense, I also find that `shell init` in conda/mamba is useful for a personal laptop setup, but for CI or Dockefiles it's simpler to go with just setting PATH.

PR #48, opened 2021-03-20
Add dataproc helper

lgruen:


vladsaveliev:
Not in the scope of this PR, but the comment to the `--description` says it has a default, however in fact the option is required and the default doesn't seem to be set in the code. I'd propose to make the parameter optional and add a default value as described.

https://github.com/populationgenomics/analysis-runner/blob/aafdd90553e44e3a9a810afe82595e2d55a8bc3a/analysis_runner/cli.py#L56-L60

lgruen:
> Not in the scope of this PR, but the comment to the `--description` says it has a default, however in fact the option is required and the default doesn't seem to be set in the code. I'd propose to make the parameter optional and add a default value as described.
> 
> https://github.com/populationgenomics/analysis-runner/blob/aafdd90553e44e3a9a810afe82595e2d55a8bc3a/analysis_runner/cli.py#L56-L60

This actually gets fixed in the follow-up PR already :). I'd rather not make it optional unless people will really dislike it when they start using the tool, as it'll add a human-readable description to the metadata that's logged with the output data. I think that would be a useful property.

vladsaveliev:
A design question: I usually put CLI scripts into a `scripts/` folder under the project root, and specify them with a `scripts=[]` parameter in `setup.py` like here: https://docs.python.org/3/distutils/setupscript.html#installing-scripts (in contrast to putting them into the package folder and specify with `entry_points`). Should we use a consistent approach in our projects? Happy to switch to your approach!

vladsaveliev:
> This actually gets fixed in the follow-up PR already :). I'd rather not make it optional unless people will really dislike it when they start using the tool, as it'll add a human-readable description to the metadata that's logged with the output data. I think that would be a useful property.

I'm just afraid we might start setting generic descriptions like "test", "test1", etc, which might be worse than the automated one :)

lgruen:
> I'm just afraid we might start setting generic descriptions like "test", "test1", etc, which might be worse than the automated one :)

Yes, that's certainly a justified concern. I thought of this similar to a GitHub PR description, but maybe that's not how people will use it?

vladsaveliev:
> Yes, that's certainly a justified concern. I thought of this similar to a GitHub PR description, but maybe that's not how people will use it?

I'd say it can be more like a GH commit message, that so often just ends up to be "fix", "fix1", "fix final" :D But let's see how it goes in practice!

vladsaveliev (approved):

PR #54, opened 2021-03-23
Add git stash to clean working directory before build

illusional:
Instructions mention to have a clean working directory, but just a small suggestion to add `git stash` to the instructions to stash the changes in the potentially dirty working directory away, then pop them back later.

Just an idea, happy to close this if it's not valuable :)

### DIFF
--- a/driver/README.md
+++ b/driver/README.md
@@ -14,7 +14,9 @@ Then run:
 ```bash
 gcloud config set project analysis-runner
 
+git stash
 IMAGE=australia-southeast1-docker.pkg.dev/analysis-runner/images/driver
 COMMIT_HASH=$(git rev-parse --short=12 HEAD)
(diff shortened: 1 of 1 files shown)


driver/README.md
```diff
@@ -14,7 +14,9 @@ Then run:
 ```bash
```
lgruen, line 14: This is actually somewhat outdated since we build the driver image as part of the Hail update workflow now and probably should never push a new image manually.

How about changing lines 11 and below to:

```markdown
The driver image gets rebuilt and pushed automatically as part of the [Hail update workflow](../.github/workflows/hail_update.yaml).
```

illusional: Yeah I completely agree, I thought we might've done this, but couldn't remember.

lgruen (approved):

PR #61, opened 2021-04-08
Append metadata entries

lgruen:
Curious to hear whether you think this is a good idea or not :)


vladsaveliev (approved):
Looks good to me. Can't see why not :) 
Though the resulting metadata file might get really cluttered with a lot of attempts on the test level.

PR #66, opened 2021-04-13
Convert click to argparse for flexibility

illusional:
I couldn't make `@click` do what we wanted, eg: parsing only a subset of `sys.argv`, but script isn't a named parameter, it's just the remainder of args. So I've reverted this back to argparse (and implemented a cheap "confirm" function).

### DIFF
--- a/analysis_runner/cli.py
+++ b/analysis_runner/cli.py
@@ -4,8 +4,8 @@
 CLI for interfacing with deployed analysis runner.
 See README.md for more information.
 """
+import argparse
 import logging
-import click
 import requests
 import google.auth
 import google.auth.transport.requests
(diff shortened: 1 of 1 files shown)


analysis_runner/cli.py
```diff
@@ -153,7 +132,64 @@ def _get_google_auth_token() -> str:
...
+    parser.add_argument('script', nargs=argparse.REMAINDER, default=[])
+
+    return parser.parse_args(args)
+
+
 if __name__ == '__main__':
-    # Disable pylint because click decorates the function in a specific way
-    # pylint: disable=no-value-for-parameter
-    main()
+    main_from_args()
```
lgruen, line 195: Shouldn't this be `main_from_args(sys.argv)` or similar? Where do the `args` come from? :)

illusional: Argparse automatically pulls from `sys.argv`.

lgruen: Got it, thanks!

illusional: I added a comment to clarify this in code.

illusional:
@vladsaveliev tests are passing.

vladsaveliev (approved):
Forgot to approve :)

PR #72, opened 2021-04-15
Pull git submodules

vladsaveliev:
Pull git submodules when cloning the repository.

### DIFF
--- a/analysis_runner/dataproc.py
+++ b/analysis_runner/dataproc.py
@@ -78,6 +78,8 @@ def hail_dataproc_job(
     main_job.command(f'git clone {git_remote} {repo_name}')
     main_job.command(f'cd {repo_name}')
     main_job.command(f'git checkout {git_sha}')
+    main_job.command(f'git submodule init')
+    main_job.command(f'git submodule update')
     main_job.command(f'cd ./{git_dir}')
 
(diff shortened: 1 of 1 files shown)


analysis_runner/dataproc.py
```diff
@@ -78,6 +78,8 @@ def hail_dataproc_job(
     main_job.command(f'git clone {git_remote} {repo_name}')
     main_job.command(f'cd {repo_name}')
     main_job.command(f'git checkout {git_sha}')
+    main_job.command(f'git submodule init')
+    main_job.command(f'git submodule update')
```
lgruen, line 82:
```suggestion
    main_job.command(f'git clone --recurse-submodules {git_remote} {repo_name}')
    main_job.command(f'cd {repo_name}')
    main_job.command(f'git checkout {git_sha}')
```

lgruen (approved):

analysis_runner/dataproc.py
vladsaveliev: I don't think it would work if the main branch doesn't yet have the submodules?

vladsaveliev: Still need `git submodule update` after checkout

PR #73, opened 2021-04-15
Pull git submodules in the server's main

vladsaveliev:
That wasn't the right place to pull submodules in the previous PR :) (though still useful to keep).

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -162,6 +162,8 @@ async def index(request):
             job.command('git checkout main')
             job.command(f'git merge-base --is-ancestor {quote(commit)} HEAD')
         job.command(f'git checkout {quote(commit)}')
+        job.command(f'git submodule init')
+        job.command(f'git submodule update')
         # Make sure the file is in a subdirectory of the repository.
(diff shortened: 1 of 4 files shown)


server/main.py
```diff
@@ -162,6 +162,8 @@ async def index(request):
             job.command('git checkout main')
             job.command(f'git merge-base --is-ancestor {quote(commit)} HEAD')
         job.command(f'git checkout {quote(commit)}')
+        job.command(f'git submodule init')
```
lgruen, line 165: Can we do a `--recurse-submodules` in the `clone` step instead?

lgruen (changes requested):

server/main.py
vladsaveliev: Don't think that would work if the submodule exists only in a branch, but not in main

vladsaveliev: Only with a subsequent `git submodule update`

vladsaveliev: Updated!

lgruen (approved):
Nice!

PR #80, opened 2021-04-20
Add other execution types

illusional:
Added R, Bash and Custom execution types by introducing `ExecutionType` enum. 
Still have to resolve which images will be used for the R and Custom types. Each image would have to contain at least `gcloud`. Currently it just resolves to the `DRIVER_IMAGE`.

 > We might consider users providing their own container, but it also means they can run arbitrary code without code review on production data. Maybe "test" could run any image, and standard must use an image from a specific artefact registry?

Any thoughts on this first draft implementation?

### DIFF
--- a/analysis_runner/cli.py
+++ b/analysis_runner/cli.py
@@ -43,6 +43,7 @@ def main(
     access_level,
     commit=None,
     repository=None,
+    execution_type=None,
 ):
     """
     Main function that drives the CLI.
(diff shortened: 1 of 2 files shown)


lgruen (changes requested):
There are two ways we could go here:

1. Explicitly have different execution modes, in a way you've structured it so far.
2. Just have one driver image that supports everything and support arbitrary commands.

(1) adds some complexity and means we need to maintain multiple images now.
(2) could lead to a bloated image.

I think as long as we only want to support standard execution environments, (1) might me a little easier? If someone really needs something more custom / bespoke, they can always kick off a Hail Batch job with their own Docker image.

What do you think?

illusional:
One image is easy! More than happy to do that :)

illusional:
Closing in favour of #82.

PR #82, opened 2021-04-21
Genericise analysis runner

illusional:
On review of https://github.com/populationgenomics/analysis-runner/pull/80, performing the following:

Server:
- still remove `python3`
- remove the inferred working directory (prev `$(basename script[0])`), in favour of an explicit `cwd`
  - This is also passed to the airtable meta
- If `script[0]` exists, (`-f`), make it executable with `chmod +x {script[0]}`

CLI:
- Determine cwd from git root
- If the `script[0]` is a file, and exists:
  - perform a check for shebang
- or see if the first element (`script[0]`) is executable (through `shutil.which`), and makes you CONFIRM if it isn't.
  - This check doesn't guarantee anything (because the container the job runs in might include some executable that your host computer doesn't), but it's safety net for misspelled scripts.

### DIFF
--- a/analysis_runner/cli.py
+++ b/analysis_runner/cli.py
@@ -70,29 +70,33 @@ def main(
     _cwd = None
 
     if ' ' in _script[0]:
+        # Some analysts will quote the whole script, eg:
+        #   analysis-runner 'myscript.py --flag --param arg'
+        # so we'll normalise this by splitting the first element.
+        # This means that your script names can't have spaces in them.
         _script = _script[0].split() + _script[1:]
 
     # false-y value catches empty list / tuple as well
     if not _script:
         _script = ['main.py']
 
-    if _script[0]:
-        if os.path.exists(_script[0]):
-            _perform_shebang_check(_script[0])
-            # if it's just the path name, eg: you call
-            #   analysis-runner myfile.py
-            # need to pre-pend "./" to execute
-            if os.path.basename(_script[0]) == _script[0]:
-                _script[0] = './' + _script[0]
-        elif not which(_script[0]):
-            # the first el of _script is not executable
-            # (at least on this computer)
-            if not confirm_choice(
-                f"The first element of the script '{_script[0]}' was not executable \n"
-                f'(or a script could not be found) on this computer. \n'
-                f'Please confirm to continue.'
-            ):
-                raise SystemExit()
+    # we can find the script, and it's a relative path (not absolute)
+    if os.path.exists(_script[0]) and not _script[0].startswith('/'):
+        _perform_shebang_check(_script[0])
+        # if it's just the path name, eg: you call
+        #   analysis-runner my_file.py
+        # need to pre-pend "./" to execute
+        if os.path.basename(_script[0]) == _script[0]:
+            _script[0] = './' + _script[0]
+    elif not which(_script[0]):
+        # the first el of _script is not executable
+        # (at least on this computer)
+        if not confirm_choice(
+            f"The first element of the script '{_script[0]}' was not executable \n"
+            f'(or a script could not be found) on this computer. \n'
+            f'Please confirm to continue.'
+        ):
+            raise SystemExit()
 
     if repository is None:
         _repository = get_repo_name_from_remote(get_git_default_remote())
@@ -158,6 +162,9 @@ def _get_google_auth_token() -> str:
 
 
 def _perform_shebang_check(script):
+    """
+    Returns None if script has shebang, otherwise raises Exception
+    """
     with open(script) as f:
         potential_shebang = f.readline()
         if potential_shebang.startswith('#!'):
@@ -171,7 +178,7 @@ def _perform_shebang_check(script):
         elif script.lower().endswith('.r') or script.lower().endswith('.rscript'):
             suggestion_shebang = '#!/usr/bin/env Rscript'
 
-        message = f'Couldn't find shebang at start of "{script}"'
+        message = f'Couldn\'t find shebang at start of "{script}"'
         if suggestion_shebang:
             message += (
                 f', consider inserting "{suggestion_shebang}" at the top of this file'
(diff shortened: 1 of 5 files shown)


analysis_runner/cli.py
```diff
@@ -63,21 +67,36 @@ def main(
...
         _script = ['main.py']
 
+    if _script[0]:
+        if os.path.exists(_script[0]):
+            _perform_shebang_check(_script[0])
+        elif not which(_script[0]):
+            # the first el of _script is not executable
+            # (at least on this computer)
+            if not confirm_choice(
+                f"The first element of the script '{_script[0]}' was not executable \n"
```
lgruen, line 86: When would this occur except for the file not being found? If it's only the latter, maybe the error message could be simplified a bit.

server/main.py
```diff
@@ -130,14 +130,15 @@ async def index(request):
...
+        cwd = params.get('cwd')
         script = params['script']
         if not script:
             raise web.HTTPBadRequest(reason='Invalid script parameter')
 
         if not isinstance(script, list):
             raise web.HTTPBadRequest(reason='Script parameter expects an array')
 
-        script_file, *script_args = script
+        # script_file, *script_args = script
```
lgruen, line 141: Remove this?

server/main.py
```diff
@@ -224,9 +218,18 @@ async def index(request):
             f'gsutil cp {METADATA_PREFIX}.json {quote(output_dir)}/metadata.json'
         )
 
+        # if
```
lgruen, line 221: Instead of `chmod`ing, could we run `/bin/sh <script>` always?

lgruen (changes requested):
Love the shebang suggestions in the error message!

analysis_runner/cli.py
illusional: This handles the case like `analysis-runner pythn3 <your-script>`, where the first element (a) is not a file; (b) is not an executable. Maybe you misspelled the executor, or some other reason, it's just there to confirm. Maybe it makes sense to add an option like (--skip-checks) to skip file-existence or executor checks like this - but can evaluate that later.

server/main.py
illusional: As discussed, `/bin/sh python-script.py` doesn't work. The file existence check is to make sure if we're running a script directly, we make it executable, otherwise if it's the executor (like `python` or `R`), it does nothing.

server/main.py
```diff
@@ -224,9 +217,17 @@ async def index(request):
             f'gsutil cp {METADATA_PREFIX}.json {quote(output_dir)}/metadata.json'
         )
 
+        job.command(
```
lgruen, line 220: Should this be dependent on the checks above? (E.g. you wouldn't want to `chmod +x /bin/bash`.)

lgruen (changes requested):

analysis_runner/cli.py
illusional: No you don't, and that's why it confirms it :) And maybe why it might make sense to add a `--skip-checks` flag.

server/main.py
illusional: That check is performed in the `cli`, but it makes sense to repeat these checks here.

analysis_runner/cli.py
```diff
@@ -63,21 +67,41 @@ def main(
     _repository = repository
     _commit_ref = commit
     _script = list(script)
+    _cwd = None
+
     if ' ' in _script[0]:
```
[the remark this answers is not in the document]
illusional: Added a comment, and a limitation to the README.

server/main.py
```diff
@@ -223,10 +216,18 @@ async def index(request):
...
             f'gsutil cp {METADATA_PREFIX}.json {quote(output_dir)}/metadata.json'
         )
+        if not script[0].startswith('/'):
+            job.command(
+                f"""\
+if [[ -f {quote(script[0])} ]] then
+    chmod +x {quote(script[0])}
+fi
+    """
+            )
```
lgruen, line 226:
```suggestion
        job.command(f'which {quote(script[0])} || chmod +x {quote(script[0])}')
```

lgruen (changes requested):

driver/Dockerfile
```diff
@@ -11,7 +11,10 @@ RUN apt-get update && apt-get install -y git wget bash bzip2 zip && \
     rm -r /var/cache/apt/* && \
     wget -qO- https://api.anaconda.org/download/conda-forge/micromamba/0.8.2/linux-64/micromamba-0.8.2-he9b6cbd_0.tar.bz2 | tar -xvj -C /usr/local bin/micromamba && \
     mkdir $MAMBA_ROOT_PREFIX && \
-    micromamba install -y --prefix $MAMBA_ROOT_PREFIX -c cpg -c bioconda -c conda-forge hail=$HAIL_VERSION analysis-runner bokeh selenium phantomjs && \
+    micromamba install -y --prefix $MAMBA_ROOT_PREFIX \
+        -c cpg -c bioconda -c conda-forge \
+        hail=$HAIL_VERSION analysis-runner bokeh selenium phantomjs \
+        r-base=4.0.3 r-essentials r-tidyverse && \
```
lgruen, line 17: Is there a particular reason we pin `r-base` to 4.0.3?

server/main.py
```diff
@@ -130,15 +130,15 @@ async def index(request):
         if not commit or commit == 'HEAD':
             raise web.HTTPBadRequest(reason='Invalid commit parameter')
 
+        # make cwd non-optional to make all old clients break
```
lgruen, line 133:
```suggestion
```

lgruen (approved):

driver/Dockerfile
illusional: That was the latest version, it looks like r-base do maintain and update some older versions.

PR #91, opened 2021-04-26
Add latest version check to CLI

illusional:
Checks https://raw.githubusercontent.com/populationgenomics/analysis-runner/main/analysis_runner/_version.py and determines if versions are the same. 

If they are different, then throw the warning:

> Your version of analysis-runner is out of date: {current_version} != {latest_version} (current vs latest). Your analysis will still be submitted, but may not work as expected. You can update the analysis-runner by running "conda install -c cpg analysis-runner={latest_version}"

There's no _greater than_ check.

### DIFF
--- a/analysis_runner/cli.py
+++ b/analysis_runner/cli.py
@@ -5,6 +5,7 @@ CLI for interfacing with deployed analysis runner.
 See README.md for more information.
 """
 import os
+import re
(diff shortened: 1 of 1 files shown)


vladsaveliev (approved):
Oh, that's awesome!

PR #93, opened 2021-04-26
Bump version: 1.6.0 → 1.7.0 -- add option to define Dataproc disk sizes

lgruen:
@KatalinaBobowik This should let you increase the HDFS capacity for the LD-pruning test.


vivbak (approved):
👍  :)

KatalinaBobowik:
> @KatalinaBobowik This should let you increase the HDFS capacity for the LD-pruning test.

Perfect, just updated it. Thanks!

PR #97, opened 2021-05-12
Use cpg-utils library instead of cloud_identity.py

lgruen:


illusional (approved):

illusional:
Will merge once it passes CI.

PR #99, opened 2021-05-12
Add web server implementation

lgruen:


illusional (approved):
Code looks good, I presume the permissions for the new web app come from the service account that it gets deployed under, and that it's deployed behind an IAP?

lgruen:
> Code looks good, I presume the permissions for the new web app come from the service account that it gets deployed under, and that it's deployed behind an IAP?

Yes, that's exactly right!

PR #106, opened 2021-05-14
Add cromwell support to analysis-runner server

illusional:
This PR adds support for Cromwell to the analysis-runner server.

- Adds an endpoint `/cromwell` which prepares a hail batch job that clones the repositories, and submits a workflow + inputs to the separately managed cromwell server. Note the following workflow options:

    ```python
    workflow_options = {
        'user_service_account_json': service_account_json,
        'google_project': project,
        'jes_gcs_root': intermediate_dir,   # gs://cpg-{dataset}-temporary/cromwell
        'final_workflow_outputs_dir': output_dir,
    }
    ```

- Refactors a lot of shared code between the regular analysis-runner `/` endpoint, including:
     - request handling, 
     - authentication of users
     - preparing the job that clones the the repository
     - standardising the metadata keys between the two requests

Currently, it does NOT add support to the CLI, that's still to come.

---

TODO:

 - [x] Add the following secrets (or determine workaround)
    - [x] `{access_level}Project` - the project ID
    - [x] `{access_level}ServiceAccount` - the JSON
 - [ ] Implement this on the CLI (not strictly necessary for initial testing)


illusional:
As suggested, instead of proxying requests through the analysis-runner, we could allow `software-team / analysis-team@popgen` to access Cromwell directly too: https://cloud.google.com/iap/docs/cloud-iap-context-aware-access-howto

server/cromwell_routes.py
```diff
@@ -0,0 +1,185 @@
...
+
+            repo = params['repo']
+            allowed_repos = server_config[dataset]['allowedRepos']
+            if repo not in allowed_repos:
+                raise web.HTTPForbidden(
+                    reason=(
+                        f'Repository "{repo}" is not one of the allowed repositories: '
+                        f'{", ".join(allowed_repos)}'
+                    )
+                )
```
lgruen, line 50: Could this also be pulled out as a shared high-level check?

server/cromwell_routes.py
```diff
@@ -0,0 +1,185 @@
...
+
+            access_level = params['accessLevel']
+            ds_config = server_config[dataset]
+            hail_token = ds_config.get(f'{access_level}Token')
+            project = ds_config.get(f'{access_level}Project')
+            service_account = ds_config.get(f'{access_level}ServiceAccount')
+            intermediate_dir = ds_config.get(f'{access_level}IntermediateDir')
```
lgruen, line 57: Wouldn't we just always put that in the temporary bucket?

server/cromwell_routes.py
```diff
@@ -0,0 +1,185 @@
...
+            )
+
+            url = run_batch_job_and_print_url(batch, wait=params.get('wait', False))
+
+            return web.Response(text=f'{url}\n')
+        except KeyError as e:
+            logging.error(e)
+            raise web.HTTPBadRequest(reason='Missing request parameter: ' + e.args[0])
+
+    @routes.get('/cromwell/engine/version')
```
lgruen, line 160: I think all of these we could just allow directly, especially if the non-destructive ones.

lgruen (changes requested):
Nice! I really like that you're refactoring the common components. But I think there's even more "validation code" we could before submitting a Hail Batch job vs submitting a Cromwell job.

server/cromwell_routes.py
illusional: Yeah, pylint actually complains about this, so worthwhile!

server/cromwell_routes.py
illusional: Is there one temporary bucket per dataset? Or one overall?

server/cromwell_routes.py
illusional: Yep, sounds great!

illusional:
Couple of merges with main and the keys are in! I think the CLI can come later, we can test more before it's built.

lgruen (approved):
This is awesome!

Not merging directly in case you want to update the PR description first.

examples/cromwell/README.md
```diff
@@ -0,0 +1,29 @@
...
+It's possible to run workflows on a managed Cromwell instance
+through the analysis-runner.
+
+The implementation for the analysis-runner CLI is still in progress,
+but you can create a POST request in the short term, for example:
+
+```shell
+curl --location \
+    --request POST 'https://server-a2pko7ameq-ts.a.run.app/cromwell' \
+    --header "Authorization: Bearer $(gcloud auth print-identity-token  )" \
```
lgruen, line 12:
```suggestion
    --header "Authorization: Bearer $(gcloud auth print-identity-token)" \
```

examples/cromwell/hello.wdl
```diff
@@ -0,0 +1,16 @@
...
+
+import "tools/echo.wdl" as E
+
+workflow hello {
+  input {
+    String? inp = "Hello, world!"
+  }
+  call E.echo as print {
+    input:
+      inp=select_first([inp, "Hello, world!"])
```
lgruen, line 11:
```suggestion
      inp = select_first([inp, "Hello, world!"])
```

server/cromwell.py
```diff
@@ -0,0 +1,247 @@
+# pylint: disable=unused-variable
```
lgruen, line 1: Was that intentionally left?

server/cromwell.py
```diff
@@ -0,0 +1,247 @@
...
+            )
+            # use the email specified by the service_account_json again
+            service_account_dict = json.loads(service_account_json)
+            if 'client_email' not in service_account_dict:
+                raise web.HTTPServerError(
+                    reason="The service_account didn't contain an entry for client_email"
+                )
+            # service_account_email = f'cromwell-{access_level}@{project}.iam.gserviceaccount.com'
+            # service_account_email = f'cromwell-runner@cromwell-305305.iam.gserviceaccount.com'
+            service_account_email = service_account_dict['client_email']
```
lgruen, line 116:
```suggestion
            service_account_email = service_account_dict.get('client_email')
            if not service_account_email:
                raise web.HTTPServerError(
                    reason="The service_account didn't contain an entry for client_email"
                )            
```

server/cromwell.py
```diff
@@ -0,0 +1,247 @@
...
+            intermediate_dir = f'gs://cpg-{dataset}-{access_level}-tmp/cromwell'
```
lgruen, line 117: This won't work for `access_level == "full"`.

server/cromwell.py
```diff
@@ -0,0 +1,247 @@
...
+
+            if not service_account_json or not intermediate_dir:
```
lgruen, line 119: Is this an obsolete check? (Shouldn't `service_account_json` be checked earlier or `get_cromwell_key` raise an exception? How could `intermediate_dir` not be set?)

server/cromwell.py
```diff
@@ -0,0 +1,247 @@
...
+                repo=repo,
+                commit=commit,
+                metadata_str=metadata,
+                output_dir=output_dir,
+                print_all_statements=False,
+            )
+            job.env('OUTPUT', output_dir)
+            # we'll need to put the commands like this, because
+            # the job.command is individually scoped, ie, the cd
+            # doesn't carry across the different job.commands
```
lgruen, line 187: I'm pretty sure that's not the case: https://github.com/populationgenomics/analysis-runner/blob/main/server/main.py#L160

server/cromwell.py
```diff
@@ -0,0 +1,247 @@
...
+                    key += f'_{idx + 1}'
+
+                inputs_cli.append(f'-F "{key}=@{value}"')
+
+            nl = '\n'
+            job.command(
+                f"""
+{nl.join(commands)}
+echo '{json.dumps(workflow_options)}' > workflow-options.json
+access_token="$(gcloud auth print-identity-token --audiences="717631777761-ec4u8pffntsekut9kef58hts126v7usl.apps.googleusercontent.com")"
```
lgruen, line 228: Those nested quotes look brittle.
```suggestion
access_token=$(gcloud auth print-identity-token --audiences=717631777761-ec4u8pffntsekut9kef58hts126v7usl.apps.googleusercontent.com)
```

server/util.py
```diff
@@ -0,0 +1,257 @@
...
+    return response.payload.data.decode('UTF-8')
+
+
+server_config = json.loads(_read_secret('server-config'))
+CROMWELL_URL = 'https://cromwell.populationgenomics.org.au'
+
+# cache the cromwell keys, we don't expect these to change
+# frequently and the server would need to be reset for other
+# server_config changes to propagate correctly.
+__cromwell_key_cache = {}
```
lgruen, line 49: I don't think that's necessary. Accessing secrets is pretty fast and this could cause some really subtle unexpected behavior.

server/util.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+def _read_secret(name: str) -> str:
+    """Reads the latest version of the given secret from Google's Secret Manager."""
+    secret_name = f'projects/analysis-runner/secrets/{name}/versions/latest'
+    response = secret_manager.access_secret_version(request={'name': secret_name})
+    return response.payload.data.decode('UTF-8')
+
+
+server_config = json.loads(_read_secret('server-config'))
```
lgruen, line 43: Let's fetch this every time, otherwise stale data will become hard to debug. This won't take long at all.

lgruen (changes requested):

server/cromwell.py
illusional: Yes, the way the function definition works inside the add_cromwell_routes throws unused-variable. There's probably a better pattern to use here, but I think it's clear what's happening.

server/cromwell.py
illusional: Yep you're right, the original change that prompted this was actually for an unrelated reason, I'll revert this bit.

server/util.py
illusional: Sounds good, will revert.

server/cromwell.py
lgruen: Yeah, it's unfortunate that one can't limit the scope of this (because the warning in general is really handy).

Also see https://github.com/PyCQA/pylint/issues/2842.

server/cromwell.py
illusional: Addressed with:

```py
if access_level == 'test':
    intermediate_dir = f'gs://cpg-{dataset}-test-tmp/cromwell'
else:
    intermediate_dir = f'gs://cpg-{dataset}-main-tmp/cromwell'
```

server/cromwell.py
illusional: Yes, it's checked earlier and intermediate_dir used to be passed in, but is now set. Removed this check.

server/util.py
```diff
@@ -0,0 +1,252 @@
...
+    return response.payload.data.decode('UTF-8')
+
+
+def get_server_config() -> dict:
+    """Get the server-config from the secret manager"""
+    return json.loads(_read_secret('server-config'))
+
+
+def get_cromwell_key(dataset, access_level):
+    """Get CROMWELL key from secrets"""
```
lgruen, line 50:
```suggestion
    """Get Cromwell key from secrets"""
```

lgruen (approved):
Excellent!

PR #116, opened 2021-05-26
Give the test users admin privileges to the test bucket

lgruen:
This is just the first step in permissions refactoring, but probably most useful to get in immediately.


pdiakumis (approved):
Thank you

PR #122, opened 2021-06-08
Copy blob attributes to response headers

lgruen:
Should fix the serving of PNGs, for example.

### DIFF
--- a/web/main.py
+++ b/web/main.py
@@ -2,7 +2,7 @@
 
 import logging
 import os
-from flask import Flask, abort, request
+from flask import Flask, abort, request, Response
 import google.cloud.storage
 from cpg_utils.cloud import is_google_group_member, email_from_id_token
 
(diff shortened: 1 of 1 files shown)


web/main.py
```diff
@@ -45,7 +45,14 @@ def handler(dataset=None, filename=None):
     if blob is None:
         abort(404)
 
-    return blob.download_as_text()
+    response = Response(blob.download_as_bytes())
```
illusional, line 48: Not that this changes the previous implementation, but this will always load files into memory before serving them. Probably not bad most of the time, but I see this details being forgotten, and trying to download a larger file causing problems.

illusional (approved):

web/main.py
lgruen: We could chunk the response here, similar to what we're doing in seqr. But I hope we're not serving hundreds of MBs this way ;)

illusional: I saw [this on StackOverflow](https://stackoverflow.com/questions/50380237/reading-really-big-blobs-without-downloading-them-in-google-cloud-streaming), about downloading to a stream, and using that as the response:

```python
def getStream(blob):
    stream = open('myStream','wb', os.O_NONBLOCK)
    streaming = blob.download_to_file(stream)
```

PR #136, opened 2021-06-16
Use IAMMember instead of IAMBinding

lgruen:
The former is authoritative, leading to different datasets overwriting each other's settings.


KatalinaBobowik (approved):
👍

PR #141, opened 2021-06-21
Cache access group members as secrets

lgruen:
To work around the incredibly slow Google Groups API.

### DIFF
--- a/stack/__main__.py
+++ b/stack/__main__.py
@@ -217,6 +217,30 @@ def main():  # pylint: disable=too-many-locals
 
     access_group = create_group(group_mail('access'))
 
+    # This secret is used as a fast cache for checking memberships in the above group.
+    access_group_cache_secret = gcp.secretmanager.Secret(
+        f'access-group-cache-secret',
+        secret_id=f'{dataset}-access-members-cache',
+        project=ANALYSIS_RUNNER_PROJECT,
+        replication=gcp.secretmanager.SecretReplicationArgs(
+            user_managed=gcp.secretmanager.SecretReplicationUserManagedArgs(
+                replicas=[
+                    gcp.secretmanager.SecretReplicationUserManagedReplicaArgs(
+                        location='australia-southeast1',
+                    ),
+                ],
+            ),
+        ),
+    )
+
+    gcp.secretmanager.SecretIamBinding(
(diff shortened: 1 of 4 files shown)


stack/__main__.py
```diff
@@ -315,6 +358,7 @@ def add_bucket_permissions(
         member=pulumi.Output.concat('serviceAccount:', WEB_SERVER_SERVICE_ACCOUNT),
     )
 
+    # TODO(@lgruen): remove this once secrets are used for checking memberships.
```
illusional, line 361: Should these be removed now?

stack/__main__.py
```diff
@@ -325,6 +369,7 @@ def add_bucket_permissions(
         roles=[gcp.cloudidentity.GroupMembershipRoleArgs(name='MEMBER')],
     )
 
+    # TODO(@lgruen): remove this once secrets are used for checking memberships.
```
illusional, line 372: This one too

illusional (commented):
2 things I don't quite understand:

1. Is this just an initial test, because I don't see a web server implementation, and there are a few TODOs.
2. I can see `--trigger-topic=access-group-cache-refresh` when deploying the function, is that topic automatically triggered on an access-group change, or are there plans to invalidate it like once a day or something (you suggest a cron job in the README).

lgruen:
> 1. Is this just an initial test, because I don't see a web server implementation, and there are a few TODOs.

Not a test, this is just a first PR in a series of changes. I didn't want to send you a few thousand lines of code at once and these can be nicely staged.

> 2. I can see `--trigger-topic=access-group-cache-refresh` when deploying the function, is that topic automatically triggered on an access-group change, or are there plans to invalidate it like once a day or something (you suggest a cron job in the README).

Cloud Scheduler sends a message to that topic every 5 minutes, which triggers the execution of this function (that's basically the Cron job).

stack/__main__.py
lgruen: Not yet, we need to update the analysis-runner / web server first.

stack/__main__.py
lgruen: Not yet, we need to update the analysis-runner / web server first.

illusional (approved):

PR #155, opened 2021-07-01
Add md5sum wdl example

pdiakumis:
Adding an md5sum task with an explicit output. You can call this as e.g. (--edited based on comments) 

```
curl --location \
    --request POST 'https://server-a2pko7ameq-ts.a.run.app/cromwell' \
    --header "Authorization: Bearer $(gcloud auth print-identity-token)" \
    --header 'Content-Type: application/json' \
    --data-raw '{
        "output": "pdiakumis/analysis-runner-test",
        "dataset": "fewgenomes",
        "repo": "analysis-runner",
        "accessLevel": "test",
        "commit": "20ab067d1d89cbfd1f3db9e49149edd7b88ccc2e",
        "inputs_dict": {
            "md5sum.prefix": "sampleB",
            "md5sum.inpf": "gs://cpg-fewgenomes-test/gvcf/batch0/NA19983.g.vcf.gz.tbi"
        },
        "input_json_paths": [],
        "workflow": "md5sum_wf.wdl",
        "dependencies": ["tools"],
        "cwd": "examples/cromwell",
        "description": "md5sum on a GVCF index"
    }'
```

@illusional the `commit` above is part of this PR so I guess it will get lost if you're looking for it in the commit history, correct? I couldn't find commit `74f91a3a246fbba261f1c2a3e06bc847c77b89cf` as shown in the README example in the `main` branch history.

### DIFF
--- a/examples/cromwell/hello.wdl
+++ b/examples/cromwell/hello.wdl
@@ -1,25 +1,16 @@
 version development
 
 import "tools/echo.wdl" as E
-import "tools/md5sum.wdl" as M
 
 workflow hello {
   input {
     String? inp = "Hello, world!"
-    File inpf
-    String prefix
   }
   call E.echo as print {
     input:
       inp = select_first([inp, "Hello, world!"])
   }
-  call M.md5sum as md5 {
-    input:
-      in_file = inpf,
-      prefix = prefix
-  }
   output {
     String out = print.out
-    File md5_res = md5.out_file
   }
 }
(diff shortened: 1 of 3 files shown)


illusional (changes requested):
True, that's a great point. As soon as this is merged in, I'll create another PR with an updated commit hash.

Can we make this a separate workflow? I'm hoping to keep the hello workflow super minimal with no inputs, but think this would be a great addition!

pdiakumis:
Ah yeah, should've put it in a separate workflow file, just couldn't come up with a good name 😅 
Will do now.

illusional (approved):

PR #156, opened 2021-07-01
Bump version: 2.0.1 → 2.1.0: Add bucket_path convenience function

lgruen:
See https://centrepopgen.slack.com/archives/C01DNQ1E109/p1625107308176700 for context.


analysis_runner/__init__.py
```diff
@@ -38,12 +42,51 @@ def output_path(filename: str, bucket_category: Optional[str] = None) -> str:
...
+    In contrast to the bucket_path function, output_path takes the "output" parameter
+    from the analysis-runner invocation into account.
+
+    Examples
+    --------
+
+    Assuming that the analysis-runner has been invoked with
+    `--dataset fewgenomes --access-level test --output 1kg_pca/v42`:
+
+    >>> from analysis_runner import output_path
```
illusional: Do these examples get automatically tested?

illusional (approved):

analysis_runner/__init__.py
illusional: If not, would be cool at some point to add these doctests to the GH action.

lgruen: Well, this particular one would rely on environment variables being set. Not sure if one could automate that easily?

illusional: Ah true, not a further consideration then :)

KatalinaBobowik (approved):
Looks great!

PR #160, opened 2021-07-01
Dataproc helper: add scopes parameter

vladsaveliev:


lgruen (approved):
You'll also want to bump the version, therefore not merging.

PR #171, opened 2021-07-12
Add dataproc label

vivbak:
Context: At the moment our billing dashboard breaks down compute costs by the `goog-dataproc-cluster-name` variable. In the analysis-runner this is defined here https://github.com/populationgenomics/analysis-runner/blob/4de2a20b29be2da14d8b5915fed57529dbb36719/analysis_runner/dataproc.py#L47

Its definition makes it difficult to correlate the cost per cluster with the task that it is performing. Also, the uniqueness of each name makes it difficult to group costs. 

Potential solutions: 

* We can create a new label e.g. `job-name` and group costs according to this parameter. Caveat: We would need to deliberately adopt this convention in other places as it is not automatically applied to each cluster like the `goog-dataproc-cluster-name` label is. 

* We can amend the current cluster-name definition here to be less unique. Caveat: I'm not too sure?


vivbak:
Update: 
With flexibility in mind, the latest revision takes in a dictionary of labels, formats it appropriately, and applies all labels to the dataproc cluster. This means that labels are defined by users as opposed to the analysis runner. Alternatively, we could take in this dictionary, as well as defining pre-set labels at this stage.

analysis_runner/dataproc.py
```diff
@@ -46,6 +49,10 @@ def hail_dataproc_job(
 
     cluster_name = f'dataproc-{uuid.uuid4().hex}'
 
+    # Format labels
+    labels_string = re.sub('"|{|}| ', '', json.dumps(labels))
+    labels_formatted = re.sub(':', '=', labels_string)
```
lgruen, line 54: Can we build this directly from `labels` without involving JSON and REs? Are certain characters not allowed in label keys or values?

examples/dataproc/main.py
```diff
@@ -19,6 +19,7 @@
     packages=['click', 'selenium'],
     init=['gs://cpg-reference/hail_dataproc/install_common.sh'],
     job_name='example',
+    labels={'compute_category': 'dataproc', 'stage': 'example_usage'},
```
lgruen, line 22: Wouldn't we want to set the category automatically, as these are all Dataproc jobs? Don't we already know that from the SKU?

lgruen (changes requested):

analysis_runner/dataproc.py
vivbak: Do you mean passing `labels` without formatting it? From what I can tell the labels need to be passed in as per the syntax in the [docs](https://cloud.google.com/dataproc/docs/guides/creating-managing-labels) i.e. 
`--labels key1=value1,key2=value2`. If we just provided the dict it looks like it will break. And yup, the characters need to be alphanumeric.

examples/dataproc/main.py
vivbak: So my line of thinking was that if we wanted to be able to break down particular costs associated with seqr (for example, annotation) then a stage=annotation label here would be useful. I was considering a scenario where `hail_dataproc_function` is used in different contexts, and therefore different breakdowns may be useful. In passing in these labels, and defining them in for example `hail-elasticsearch-pipelines` as opposed to here we can breakdown the costs with whatever level of granularity we like, and change flexibly according to our use case. Perhaps I can set this variable to be optional and automatically set these categories if the user doesn't provide them? Or is there a better way to achieve this?

analysis_runner/dataproc.py
lgruen: The rules for labels are actually a bit more complicated: https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements

I'd suggest adding a link to that documentation to the parameter. We don't need to duplicate the checks in the code here, as the cluster creation will fail anyway if the labels aren't well-formed.

> Do you mean passing labels without formatting it?

No, I just meant that converting to JSON and then modifying the resulting string is a bit of a roundabout way. How about this:

```python
labels_formatted = ','.join(f'{key}={value}' for key, value in labels.items())
```
```

File: examples/dataproc/main.py (hunk 2 above)
Comment (lgruen) [reply examples/dataproc/main.py:22]: Completely agree that we want to allow additional, optional labels. But isn't the category here fixed / redundant?

**Review by lgruen — CHANGES_REQUESTED** [review]

File: examples/dataproc/main.py (hunk 2 above)
Comment (vivbak) [reply examples/dataproc/main.py:22]: Oh, this is an example script that was created elsewhere, and because the labels were set as compulsory in `dataproc.py` I had to pass in values. This doesn't change the functionality of the `dataproc.py` script. Happy to switch this out to an optional variable though in `dataproc.py` and then this `examples/dataproc/main.py` can remain as it was before.

File: analysis_runner/dataproc.py (hunk 1 above)
Comment (vivbak) [reply analysis_runner/dataproc.py:54]: > I'd suggest adding a link to that documentation to the parameter. We don't need to duplicate the checks in the code here, as the cluster creation will fail anyway if the labels aren't well-formed.

Will do! 

> No, I just meant that converting to JSON and then modifying the resulting string is a bit of a roundabout way. How about this:

Ah, I understand. That's much better than what I had, thank you!

**Review by vivbak — COMMENTED** [review]
Replied in Files Changed -- for some reason I couldn't reply here.

File: analysis_runner/dataproc.py [hunk 3]
```diff
@@ -47,6 +48,12 @@ def hail_dataproc_job(
 
     cluster_name = f'dataproc-{uuid.uuid4().hex}'
 
+    # Add default labels to passed labels.
+    labels.update({'compute_category': 'dataproc'})
```
lgruen, line 52: Does this work when `labels` is `None`?

analysis_runner/dataproc.py
```diff
@@ -47,6 +48,12 @@ def hail_dataproc_job(
...
+
+    # Format labels
```
lgruen, line 54: No need for comments that don't add extra information to the code.

```suggestion
```

examples/dataproc/main.py
lgruen: Yes, I think it would be good to keep the example as before, for simplicity.

analysis_runner/dataproc.py
lgruen: Is the plan to also add this a default `compute_category` to Cromwell workflows?

lgruen (changes requested):

analysis_runner/dataproc.py
vivbak: Yes to both :)

lgruen: Are you sure? Wouldn't this call `None.update`?

lgruen (changes requested):

analysis_runner/dataproc.py
vivbak: ah great point! I conflated `None` and `{}` in my head.

lgruen (approved):

PR #173, opened 2021-07-20
Resolve group memberships transitively

lgruen:
Also switch to async Python for the parts that can be parallelized.

Addresses part of #174.

### DIFF
--- a/access_group_cache/main.py
+++ b/access_group_cache/main.py
@@ -121,9 +121,9 @@ async def _get_dataset_access_group_members(
 ) -> Dict[str, List[str]]:
     access_token = await _get_service_account_access_token()
 
-    group_names = [
+    group_names = (
         f'{dataset}-access@populationgenomics.org.au' for dataset in datasets
-    ]
+    )
     results = await asyncio.gather(
         *(
             _transitive_group_members(access_token, group_name)
@@ -145,10 +145,6 @@ def access_group_cache(unused_data, unused_context):
 
     for dataset in config:
         secret_value = ','.join(sorted(group_members[dataset]))
-
-        # TODO
-        print(f'{dataset}: {secret_value}')
-        continue
 
         # Check whether the current secret version is up-to-date.
         secret_name = f'{dataset}-access-members-cache'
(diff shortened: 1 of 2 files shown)


access_group_cache/main.py
```diff
@@ -40,60 +28,137 @@ def _read_secret(name: str) -> Optional[str]:
...
 
-def access_group_cache(unused_data, unused_context):
-    """Main entry point."""
-    config = json.loads(_read_secret('server-config'))
-    # Given that each group takes multiple seconds to process (!), we're running them
-    # concurrently. We're using processes, as google-api-python-client is not thread-safe:
-    # https://googleapis.github.io/google-api-python-client/docs/thread_safety.html
-    with multiprocessing.Pool(20) as pool:
-        pool.map(_process_dataset_group, config)
+        response = secret_manager.add_secret_version(
```
illusional, line 157: Is there a way to disable old versions with this API?

illusional (approved):

access_group_cache/main.py
lgruen: Yes, there are actually three different actions you can perform: disable, destroy, and delete: https://googleapis.dev/python/secretmanager/latest/secretmanager_v1/secret_manager_service.html#google.cloud.secretmanager_v1.services.secret_manager_service.SecretManagerServiceAsyncClient.delete_secret

Are you worried that we might accumulate too many versions over time?

illusional: I have a similar block in a script, and wondering if you can create and disable all in one call (like you can with the UI).

lgruen: I couldn't find anything directly in the API, therefore tried to address this in https://github.com/populationgenomics/cpg-utils/pull/5.

lgruen: I'll merge this for now, as the follow-up PR for using `cpg-utils` will also require using Cloud Run instead of Cloud Functions: a Cloud Function can only depend on pip packages, but we use Anaconda. It'll hopefully be easier to review the respective change by staging the PRs.

PR #176, opened 2021-07-21
Use cpg-utils functions for reading / writing secrets

lgruen:
Required moving from a Cloud Function deployment to Cloud Run, as our `cpg-utils` package is only available in Anaconda.

Depends on https://github.com/populationgenomics/cpg-utils/pull/6.


illusional (approved):
That's seriously lame that we have to create a container and move everything to cloud run because of a conda package

PR #184, opened 2021-07-23
Add sample-metadata api to driver container

illusional:
- Fixed sample-metadata conda deploy (in 1.0.2)
- Add sample-metadata to make it simpler for analysis tasks to update sample-metadata database.

### DIFF
--- a/driver/Dockerfile
+++ b/driver/Dockerfile
@@ -13,7 +13,7 @@ RUN apt-get update && apt-get install -y git wget bash bzip2 zip && \
     mkdir $MAMBA_ROOT_PREFIX && \
     micromamba install -y --prefix $MAMBA_ROOT_PREFIX \
         -c cpg -c bioconda -c conda-forge \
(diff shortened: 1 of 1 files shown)


driver/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apt-get update && apt-get install -y git wget bash bzip2 zip && \
     mkdir $MAMBA_ROOT_PREFIX && \
     micromamba install -y --prefix $MAMBA_ROOT_PREFIX \
         -c cpg -c bioconda -c conda-forge \
-        hail=$HAIL_VERSION analysis-runner bokeh selenium phantomjs \
+        hail=$HAIL_VERSION analysis-runner sample-metadata bokeh selenium phantomjs \
```
lgruen, line 16: Just double-checking whether we'd want to pin this version?

lgruen (approved):

driver/Dockerfile
illusional: Yeah good idea, I'll pin to start until there's more stability.

PR #186, opened 2021-07-26
Gives hail analysis-runner accounts access to sample-metadata server

illusional:


stack/__main__.py
```diff
@@ -787,6 +788,17 @@ def find_service_account(kind: str, access_level: str) -> Optional[str]:
             member=f'serviceAccount:{ANALYSIS_RUNNER_SERVICE_ACCOUNT}',
         )
 
+    for access_level, service_account in service_accounts['hail']:
+        # Give access to sample-metadata cloud run service
+        gcp.cloudrun.IamMember(
+            f'samplemetadata-service-account-{access_level}-accessor',
```
lgruen, line 794: Doesn't make a difference semantically, just checking if this was a typo.

```suggestion
            f'sample-metadata-service-account-{access_level}-accessor',
```

stack/__main__.py
```diff
@@ -787,6 +788,17 @@ def find_service_account(kind: str, access_level: str) -> Optional[str]:
...
+            location='australia-southeast1',
```
lgruen, line 795:
```suggestion
            location=REGION,
```

lgruen (changes requested):

stack/__main__.py
illusional: Just going for the one block, but this is better.

stack/__main__.py
```diff
@@ -787,6 +788,17 @@ def find_service_account(kind: str, access_level: str) -> Optional[str]:
             member=f'serviceAccount:{ANALYSIS_RUNNER_SERVICE_ACCOUNT}',
         )
 
+    for access_level, service_account in service_accounts['hail']:
```
lgruen, line 791: Would Dataproc or Cromwell service accounts need access,too?

lgruen (changes requested):

stack/__main__.py
illusional: Maybe, it's not considered good practice to update external state from within a WDL workflow, though it is convenient. More plausible from Hail. Should I add both as "just in case".

stack/__main__.py
```diff
@@ -787,6 +788,17 @@ def find_service_account(kind: str, access_level: str) -> Optional[str]:
             member=f'serviceAccount:{ANALYSIS_RUNNER_SERVICE_ACCOUNT}',
         )
 
+    for kind, access_level, service_account in service_accounts_gen():
+        # Give hail / dataproc / cromwell access to sample-metadata cloud run service
+        gcp.cloudrun.IamMember(
+            f'sample-metadata-service-account-{kind}-{access_level}-accessor',
```
lgruen, line 794:
```suggestion
            f'sample-metadata-service-account-{kind}-{access_level}-invoker',
```

lgruen (approved):

PR #187, opened 2021-07-27
Bump version: 2.4.0 → 2.5.0 to allow specifying cwd on CLI

illusional:
- Add cwd as option on CLI
- Fix 'which' and is executable check.


lgruen (approved):
Do you want to add a bump version commit for this to be released?

PR #196, opened 2021-08-09
SM Dependency 1.0.2 -> 1.0.9

vivbak:


illusional (approved):
I think in future we should make it a build arg like Hail, and then [add an input](https://github.com/populationgenomics/analysis-runner/blob/968f3b2623813b4c43cd7970ae95c47d986e471b/.github/workflows/hail_update.yaml#L7-L10) to the GH action.

But for now this is good. You'll need to run the deploy workflow to release these changes :)

PR #227, opened 2021-08-23
Add amp and perth dataset configs

illusional:


stack/Pulumi.perth-neuro.yaml
```diff
@@ -0,0 +1,11 @@
+encryptionsalt: v1:SPQXlVggjxw=:v1:X1sTpViNuyK+Wcom:3GANI8gZOKtk0gG7BklsyHeNU5uVLw==
+config:
+  datasets:customer_id: C010ys3gt
+  datasets:depends_on: '["seqr"]'
```
lgruen, line 4: This is backwards -- seqr depends on perth-neuro, not the other way around.

lgruen (changes requested):

stack/Pulumi.perth-neuro.yaml
illusional: Ahh, my initial reading was right.

lgruen (approved):
Awesome!

PR #228, opened 2021-08-23
Allow gatk-sv repo

pdiakumis:
For running code directly from <https://github.com/populationgenomics/gatk-sv>, instead of working with git submodules.

### DIFF
--- a/tokens/main.py
+++ b/tokens/main.py
@@ -11,10 +11,16 @@ from google.cloud import secretmanager
 ALLOWED_REPOS = {
     'acute-care': [],
     'ancestry': ['ancestry'],
-    'fewgenomes': ['analysis-runner', 'fewgenomes', 'joint-calling', 'sv-workflows'],
+    'fewgenomes': [
+        'analysis-runner',
+        'fewgenomes',
+        'joint-calling',
+        'sv-workflows',
+        'gatk-sv',
+    ],
     'seqr': ['hail-elasticsearch-pipelines'],
     'thousand-genomes': [],
-    'tob-wgs': ['ancestry', 'joint-calling', 'tob-wgs', 'sv-workflows'],
+    'tob-wgs': ['ancestry', 'joint-calling', 'tob-wgs', 'sv-workflows', 'gatk-sv'],
 }
 
 GCP_PROJECT = 'analysis-runner'


tokens/main.py
```diff
@@ -11,10 +11,16 @@
 ALLOWED_REPOS = {
     'acute-care': [],
     'ancestry': ['ancestry'],
-    'fewgenomes': ['analysis-runner', 'fewgenomes', 'joint-calling', 'sv-workflows'],
+    'fewgenomes': [
+        'analysis-runner',
+        'fewgenomes',
+        'joint-calling',
+        'sv-workflows',
```
lgruen, line 18: Why do we have `sv-workflows` _and_ `gatk-sv` here?

And is the plan to integrate this with `joint-calling`?

Do you have a GitHub issue / Google doc that describes the overall design / desired outcome?

lgruen (changes requested):

tokens/main.py
pdiakumis: The GATK-SV Batch mode steps/subworkflows are described here:
<https://github.com/broadinstitute/gatk-sv#cohort-mode>
I initially set out to execute them through a git submodule in `sv-workflows`. But it seems simpler (might be wrong) to work directly with the fork. Would you suggest an alternative way of executing/testing those subworkflows?

lgruen: What I'd like to avoid is confusion about where our workflows that we run reside. Having one clear repo (potentially with submodules where necessary) would help with that.

@vladsaveliev recently has added options to skip certain stages of the seqr pipeline, which is super helpful for testing of individual components / subworkflows like you describe.

Would that be a possible approach here? E.g. could we aim for GATK-SV becoming part of our "standard pipeline", but if desired, you can run only a subset of stages, configured through options / parameters?

This might also warrant a discussion with @vladsaveliev, as he's thinking about merging the joint calling pipeline and the seqr pipeline anyway. Integrating GATK-SV into both of these is crucial for many upcoming projects.

pdiakumis: Okay, I'll close this for now and catch up with Vlad during the week. I'll continue testing the subworkflows in `sv-workflows`.

PR #230, opened 2021-08-25
README: use fewgenomes, not tob-wgs

pdiakumis:
Just a fix in the README - analysis-runner doesn't have access to tob-wgs, so changed it to fewgenomes.
A couple linting errors in the actions (from unrelated code) have popped up, so not sure if I should fix those here as well..


illusional (approved):
Thanks @pdiakumis!

PR #232, opened 2021-08-31
Add sample-metadata access permissions

illusional:
Adds 4 groups and secrets for production-read, production-write, test-read, test-write, and implementing the following table:

image

I believe the depends_on relationships should be transitively carried when the groups are resolved.

### DIFF
--- a/access_group_cache/main.py
+++ b/access_group_cache/main.py
@@ -121,9 +121,6 @@ def index():
         cpg_utils.cloud.read_secret(ANALYSIS_RUNNER_PROJECT_ID, 'server-config')
     )
 
-    # key = f'sample-metadata-{env}-{rs}'
-    # secret_id = f'{dataset}-{key}-members-cache'
-
     group_types = [
         'access',
         'web-access',
--- a/access_group_cache/main.py
+++ b/access_group_cache/main.py
@@ -131,7 +131,7 @@ def index():
     # add SM group types
     group_types.extend(
         f'sample-metadata-{env}-{rs}'
-        for env in ('production', 'test')
+        for env in ('main', 'test')
         for rs in ('read', 'write')
     )
 
--- a/access_group_cache/main.py
+++ b/access_group_cache/main.py
@@ -121,10 +121,27 @@ def index():
         cpg_utils.cloud.read_secret(ANALYSIS_RUNNER_PROJECT_ID, 'server-config')
     )
 
+    # key = f'sample-metadata-{env}-{rs}'
+    # secret_id = f'{dataset}-{key}-members-cache'
+
+    group_types = [
+        'access',
+        'web-access',
+        'test',
+        'standard',
+        'full',
+    ]
+    # add SM group types
+    group_types.extend(
+        f'sample-metadata-{env}-{rs}'
+        for env in ('production', 'test')
+        for rs in ('read', 'write')
+    )
+
     groups = []
     dataset_by_group = {}
     for dataset in config:
-        for group_type in 'access', 'web-access', 'test', 'standard', 'full':
+        for group_type in group_types:
             group = f'{dataset}-{group_type}'
             groups.append(group)
             dataset_by_group[group] = dataset
--- a/stack/__main__.py
+++ b/stack/__main__.py
@@ -247,7 +247,6 @@ def main():  # pylint: disable=too-many-locals
         )
 
     def add_access_group_cache_as_secret_member(secret, resource_prefix: str):
-
         gcp.secretmanager.SecretIamMember(
             f'{resource_prefix}-group-cache-secret-accessor',
             secret_id=secret.id,
--- a/stack/__main__.py
+++ b/stack/__main__.py
@@ -344,7 +344,7 @@ def main():  # pylint: disable=too-many-locals
 
             gcp.cloudidentity.GroupMembership(
                 f'sample-metadata-group-cache-{env}-{rs}-access-level-group-membership',
-                group=group.id,
+                group=group,
                 preferred_member_key=gcp.cloudidentity.GroupMembershipPreferredMemberKeyArgs(
                     id=SAMPLE_METADATA_API_SERVICE_ACCOUNT
                 ),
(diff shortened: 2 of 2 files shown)


access_group_cache/main.py
```diff
@@ -121,10 +121,24 @@ def index():
...
+        'access',
+        'web-access',
+        'test',
+        'standard',
+        'full',
+    ]
+    # add SM group types
+    group_types.extend(
+        f'sample-metadata-{env}-{rs}'
+        for env in ('production', 'test')
```
lgruen, line 134: Should we name this `main` and `test`, for symmetry with the buckets?

The "opposite" of _prod_ would be _dev_, which is not what you mean here, IIUC. I.e. test projects like `seqr-test` would still be in the production instance.

stack/__main__.py
```diff
@@ -331,13 +331,76 @@ def create_group(mail: str) -> gcp.cloudidentity.Group:
...
 
-    for group_prefix in ACCESS_LEVELS:
-        gcp.secretmanager.SecretIamMember(
-            f'sample-metadata-api-{group_prefix}-access-group-cache-secret-accessor',
-            secret_id=access_group_cache_secrets[group_prefix].id,
-            role='roles/secretmanager.secretAccessor',
-            member=f'serviceAccount:{SAMPLE_METADATA_API_SERVICE_ACCOUNT}',
-        )
+    # Sample metadata access
+    #   - 4 secrets, production-read, production-write, development-read, development-write
```
lgruen, line 335: Obsolete comment ("development").

stack/__main__.py
```diff
@@ -331,13 +331,76 @@ def create_group(mail: str) -> gcp.cloudidentity.Group:
...
+    for env in ('production', 'test'):
+        for rs in ('read', 'write'):
+            key = f'sample-metadata-{env}-{rs}'
+            secret_id = f'{dataset}-{key}-members-cache'
+
+            group = create_group(group_mail(dataset, key))
+            sm_groups[f'{env}-{rs}'] = group
+
+            gcp.cloudidentity.GroupMembership(
+                f'sample-metadata-group-cache-{env}-{rs}-access-level-group-membership',
```
lgruen, line 346: This is not directly tied to an access level anymore, maybe we can remove that from the name?

stack/__main__.py
```diff
@@ -331,13 +331,76 @@ def create_group(mail: str) -> gcp.cloudidentity.Group:
...
+                group=group,
+                preferred_member_key=gcp.cloudidentity.GroupMembershipPreferredMemberKeyArgs(
+                    id=SAMPLE_METADATA_API_SERVICE_ACCOUNT
+                ),
+                roles=[gcp.cloudidentity.GroupMembershipRoleArgs(name='MEMBER')],
+                opts=pulumi.resource.ResourceOptions(depends_on=[cloudidentity]),
+            )
+
+            gcp.secretmanager.Secret(
```
lgruen, line 355: In order for this secret to get populated, the `ACCESS_GROUP_CACHE_SERVICE_ACCOUNT` will need to be a member of the group (to see members) as well have the necessary permissions to read and write this secret.

stack/__main__.py
```diff
@@ -331,13 +331,76 @@ def create_group(mail: str) -> gcp.cloudidentity.Group:
...
+            )
+
+            gcp.secretmanager.SecretIamMember(
+                f'sample-metadata-api-{env}-{rs}-access-group-cache-secret-accessor',
+                secret_id=secret_id,
+                role='roles/secretmanager.secretAccessor',
+                member=f'serviceAccount:{SAMPLE_METADATA_API_SERVICE_ACCOUNT}',
+            )
+
+    sm_access_levels = [
```
lgruen, line 377: I think this structure warrants a comment. Or maybe you could use a `NamedTuple` to make this clearer?

lgruen (changes requested):

stack/__main__.py
illusional: Replaced with main, test

stack/__main__.py
illusional: Yep, makes a lot of sense. I've fixed a lot the names across sample-metadata permissions here.

stack/__main__.py
illusional: Thanks, added `accessor` and `version-manager` to mirror the previous pattern.

stack/__main__.py
illusional: Yep, good idea. I've added a `namedtuple` and used `kwargs`, it's a bit more verbose but more clear.

stack/__main__.py
```diff
@@ -331,13 +335,98 @@ def create_group(mail: str) -> gcp.cloudidentity.Group:
...
+                f'{key}-group-cache-secret-version-manager',
+                secret_id=secret_id,
+                role='roles/secretmanager.secretVersionManager',
+                member=f'serviceAccount:{ACCESS_GROUP_CACHE_SERVICE_ACCOUNT}',
+            )
+
+    # Declare access to sample-metadata API of format ({env}-{read,write})
+    sm_access_levels: List[SampleMetadataAccessorMembership] = [
+        SampleMetadataAccessorMembership(
+            name='person',
```
lgruen, line 397: (For consistency with other documentation.)

```suggestion
            name='human',
```

access_group_cache/main.py
illusional: Yes, totally agree! Finished this rename across both files.

stack/__main__.py
```diff
@@ -29,6 +29,10 @@
 )
 ACCESS_LEVELS = ('test', 'standard', 'full')
 
+SampleMetadataAccessorMembership = namedtuple(
+    'SampleMetadataAccessorMembership', ['name', 'service_account', 'permissions']
```
lgruen, line 33: The second entry is really a group, not just a single account, right?

stack/__main__.py
```diff
@@ -331,13 +335,98 @@ def create_group(mail: str) -> gcp.cloudidentity.Group:
...
+                role='roles/secretmanager.secretAccessor',
+                member=f'serviceAccount:{ACCESS_GROUP_CACHE_SERVICE_ACCOUNT}',
+            )
+
+            gcp.secretmanager.SecretIamMember(
+                f'{key}-group-cache-secret-version-manager',
+                secret_id=secret_id,
+                role='roles/secretmanager.secretVersionManager',
+                member=f'serviceAccount:{ACCESS_GROUP_CACHE_SERVICE_ACCOUNT}',
+            )
```
lgruen, line 392: We have a similar sequence in other places in this file. Not sure whether it's easy to factor out, but might be worth considering.

lgruen (approved):

stack/__main__.py
illusional: I've renamed this `member_key` to more closely align with Pulumi.

stack/__main__.py
```diff
@@ -223,6 +229,39 @@ def create_group(mail: str) -> gcp.cloudidentity.Group:
...
+                        ),
+                    ],
+                ),
+            ),
+            opts=pulumi.resource.ResourceOptions(depends_on=[secretmanager]),
+            **kwargs,
+        )
+
+    def add_access_group_cache_as_secret_member(secret, resource_prefix: str):
+
```
lgruen, line 250:
```suggestion
```

lgruen (approved):
Very nice, thanks!

PR #237, opened 2021-09-06
Add cromwell labels

vivbak:
Add custom labels to cromwell instance as well as standard compute-category label.


analysis_runner/cromwell.py
```diff
@@ -151,6 +160,11 @@ def run_cromwell(
                 workflow_input_prefix + k: v for k, v in _inputs_dict.items()
             }
 
+    if labels:
+        _labels = json.loads(labels)
+    if labels is None:
+        _labels = None
```
illusional, line 166:
```suggestion
    _labels = None
    if labels:
        _labels = json.loads(labels)
```

analysis_runner/dataproc.py
```diff
@@ -50,7 +50,7 @@ def hail_dataproc_job(
 
     if labels is None:
         labels = {}
-    labels.update({'compute_category': 'dataproc'})
+    labels.update({'compute-category': 'dataproc'})
```
illusional, line 53:
```suggestion
    labels['compute-category'] = 'dataproc'
```

server/cromwell.py
```diff
@@ -199,12 +200,18 @@ async def cromwell(request):
                 )
 
             cromwell_post_url = CROMWELL_URL + '/api/workflows/v1'
+
+            google_labels = {'compute-category': 'cromwell'}
+            if labels:
+                google_labels.update(labels)
```
illusional, line 206: It's probably fine, but this does allow the user to override `compute-category` to be whatever they want. You could reorder this if that's unintended behaviour.

illusional (approved):
Awesome work!

server/cromwell.py
vivbak: Great point, thanks! I've swapped the order.

PR #243, opened 2021-09-08
Dataproc helper small fix: use equal sign for --labels=

vladsaveliev:
Using space to separate the value for the `--label` option unfortunately breaks hailctl: https://batch.hail.populationgenomics.org.au/batches/4941/jobs/1

FYI @vivbak


vivbak (approved):
Oh, sorry about that Vlad!

PR #245, opened 2021-09-10
Add ref-db stack

illusional:
Notably: depends_on: 
- tob-wgs
- mgrb
- thousand-genomes


lgruen (commented):
This looks good, but let's wait for a decision on the name first :).

PR #246, opened 2021-09-10
Add functions to start, submit and stop Dataproc clusters separately

vladsaveliev:
Being able to start a cluster once and submit multiple jobs consequently would benefit a lot the joint-calling pipeline a lot, as it currently submits lots of hail query scripts, each of them involving creating almost identical clusters.

### DIFF
--- a/analysis_runner/dataproc.py
+++ b/analysis_runner/dataproc.py
@@ -44,7 +44,7 @@ class DataprocCluster:
         script: str,
         job_name: Optional[str] = None,
         pyfiles: Optional[List[str]] = None,
-    ):
+    ) -> hb.batch.job.Job:
         """
         Create a job that submits the `script` to the cluster
         """
@@ -57,6 +57,7 @@ class DataprocCluster:
         )
         job.depends_on(self.start_job)
         self.stop_job.depends_on(job)
+        return job
 
 
 def hail_dataproc(
--- a/analysis_runner/dataproc.py
+++ b/analysis_runner/dataproc.py
@@ -2,7 +2,6 @@
 
 import os
 import uuid
-import contextlib
 from typing import Optional, List, Dict, Tuple
 import hailtop.batch as hb
 from analysis_runner.git import (
@@ -40,7 +39,7 @@ class DataprocCluster:
         self.stop_job = stop_job
         self.cluster_name = cluster_name
 
-    def submit(
+    def add_job(
         self,
         script: str,
         job_name: Optional[str] = None,
@@ -60,7 +59,6 @@ class DataprocCluster:
         self.stop_job.depends_on(submit_job)
 
 
-@contextlib.contextmanager
 def hail_dataproc(
     batch: hb.Batch,
     *,
@@ -78,7 +76,7 @@ def hail_dataproc(
     job_name: Optional[str] = None,
     scopes: Optional[List[str]] = None,
     labels: Optional[Dict[str, str]] = None,
-) -> hb.batch.job.Job:
+) -> DataprocCluster:
     """
     Adds jobs to the batch that start and stop a Dataproc cluster, and returns
     a DataprocCluster object with a submit() method that allows to add jobs into
@@ -115,11 +113,9 @@ def hail_dataproc(
     )
     stop_job.depends_on(start_job)
 
-    yield DataprocCluster(
+    return DataprocCluster(
         batch=batch, start_job=start_job, stop_job=stop_job, cluster_name=cluster_name
     )
-
-    return stop_job
 
 
 def hail_dataproc_job(
--- a/analysis_runner/dataproc.py
+++ b/analysis_runner/dataproc.py
@@ -2,7 +2,7 @@
 
 import os
 import uuid
-from typing import Optional, List, Dict
+from typing import Optional, List, Dict, Tuple
 import hailtop.batch as hb
 from analysis_runner.git import (
     get_git_default_remote,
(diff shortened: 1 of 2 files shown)


lgruen (changes requested):
While I like the additional flexibility of this more granular API, it's also more error-prone. In particular, the previous API guaranteed that the cluster was shut down, whereas now it's up to the caller to call `stop_cluster` correctly.

Can we enforce correct usage by using a `with ... as ...:` pattern or similar?

While I understand that recreating the cluster multiple times is annoying, does it really save that much time overall? Wouldn't most jobs that we run take on the order of hours, whereas cluster startup takes a few minutes? In other words, is it really worth optimizing this, at the expense of a more difficult to use API?

vladsaveliev:
> While I like the additional flexibility of this more granular API, it's also more error-prone. In particular, the previous API guaranteed that the cluster was shut down, whereas now it's up to the caller to call stop_cluster correctly.

The `hail_dataproc_job` function is still the preferable way. The new functions are primarily for the case when you know what you are doing, and I imagine them to be used just from the `joint-calling` pipeline.

> While I understand that recreating the cluster multiple times is annoying, does it really save that much time overall? Wouldn't most jobs that we run take on the order of hours, whereas cluster startup takes a few minutes? In other words, is it really worth optimizing this, at the expense of a more difficult to use API?

That's true for the production runs. But most of the runs I make are test runs on a toy dataset, where most of the jobs run in minutes, the cluster starting jobs are the real bottleneck. Slow test runs significantly slow down the development.

> Can we enforce correct usage by using a with ... as ...: pattern or similar?

Good idea! I'll try.

vladsaveliev:
Actually, no need to `with/as` because the dependency tree is defined later when `batch.run()` is called. So can organise it simply as follows:

```python
cluster = dataproc.hail_dataproc(
    batch,
    max_age='1h',
    init=['gs://cpg-reference/hail_dataproc/install_common.sh'],
)
cluster.add_job('query_script1.py', job_name='example 1')
cluster.add_job('query_script2.py', job_name='example 2')
```

Where `cluster` is a `DataprocCluster` object with a `add_job()` method that inserts the new job between start and stop into the dependency tree:

```python
job.depends_on(self.start_job)
self.stop_job.depends_on(job)
```

We might even replace use the `dataproc.py` module with a singleton instead of a `DataprocCluster` object, and keep the old interface `dataproc.hail_dataproc_job()` functional and avoid breaking compatibility. If the cluster parameters change, we start a new cluster. Not sure if this approach would be transparent enough though. What do you think?

analysis_runner/dataproc.py
```diff
@@ -21,6 +22,102 @@
...
+        start_job: hb.batch.job.Job,
+        stop_job: hb.batch.job.Job,
+        cluster_name: str,
+    ):
+        self.batch = batch
+        self.start_job = start_job
+        self.stop_job = stop_job
+        self.cluster_name = cluster_name
+
+    def add_job(
```
lgruen, line 42: Should this return the job object, so one can define dependencies between different jobs that are submitted to the same cluster?

analysis_runner/dataproc.py
```diff
@@ -21,6 +22,102 @@
...
+    """
+
+    def __init__(
+        self,
+        batch: hb.Batch,
+        start_job: hb.batch.job.Job,
+        stop_job: hb.batch.job.Job,
+        cluster_name: str,
+    ):
+        self.batch = batch
```
lgruen, line 37: Should these be private class members, i.e. `_batch` or similar?

analysis_runner/dataproc.py
```diff
@@ -21,6 +22,102 @@
...
+    requester_pays_allow_all: bool = True,
+    depends_on: Optional[List[hb.batch.job.Job]] = None,
+    job_name: Optional[str] = None,
+    scopes: Optional[List[str]] = None,
+    labels: Optional[Dict[str, str]] = None,
+) -> DataprocCluster:
+    """
+    Adds jobs to the batch that start and stop a Dataproc cluster, and returns
+    a DataprocCluster object with a submit() method that allows to add jobs into
+    this cluster that submit Query scripts.
```
lgruen, line 83:
```suggestion
    a DataprocCluster object with an add_job() method to submit Hail Query scripts.
```

analysis_runner/dataproc.py
```diff
@@ -21,6 +22,102 @@
...
+
+    See the `hailctl` tool for information on the keyword parameters.
+
+    depends_on can be used to enforce dependencies for the new job.
```
lgruen, line 87:
```suggestion
    depends_on can be used to enforce dependencies for the cluster startup.
```

examples/dataproc/main.py
```diff
@@ -12,14 +12,25 @@
...
+
+
+# dataproc.hail_dataproc_job(
+#     batch,
+#     'query.py',
+#     max_age='1h',
+#     packages=['click', 'selenium'],
+#     init=['gs://cpg-reference/hail_dataproc/install_common.sh'],
+#     job_name='example',
+# )
```
lgruen, line 33:
```suggestion

cluster.add_job('query.py', job_name='example')
```

lgruen (changes requested):
Nice!

analysis_runner/dataproc.py
vladsaveliev: Good catch. Done!

analysis_runner/dataproc.py
vladsaveliev: I wanted to be explicit that we don't actually submit Query scripts here. We add a job will submit.

vladsaveliev: But fixed the method name!

analysis_runner/dataproc.py
```diff
@@ -32,32 +74,116 @@ def hail_dataproc_job(
...
-    cluster_name = f'dataproc-{uuid.uuid4().hex}'
+    See the `hailctl` tool for information on the keyword parameters.
+    """
+    job_name_prefix = f'{job_name}: s' if job_name else 'S'
+    job_name = f'{job_name_prefix}tart Dataproc cluster'
+    if cluster_title:
+        job_name += f' "{cluster_title}"'
+
+    cluster_id = cluster_title or 'dataproc'
+    cluster_id = re.sub(r'[-\s]+', '-', cluster_id.lower())
```
lgruen, line 178:
```suggestion
    cluster_id = re.sub(r'^[a-zA-Z0-9]+', '-', cluster_id.lower())
```

analysis_runner/dataproc.py
```diff
@@ -21,9 +23,49 @@
...
+        start_job: hb.batch.job.Job,
+        stop_job: hb.batch.job.Job,
+        cluster_id: str,
+        cluster_title: Optional[str] = None,
+    ):
+        self._batch = batch
+        self._start_job = start_job
+        self._stop_job = stop_job
+        self._cluster_id = cluster_id
+        self.cluster_title = cluster_title
```
lgruen, line 43: Should this also be private? And how about `cluster_name` instead of `cluster_title`?

```suggestion
        self._cluster_name = cluster_name
```

lgruen (approved):

PR #249, opened 2021-09-12
Bump version: 2.6.0 → 2.6.1: Fix regex character class inversion

lgruen:


vladsaveliev (approved):
Argh, sorry for the regex bug! Shouldn't have rushed.

lgruen:
> Argh, sorry for the regex bug! Shouldn't have rushed.

No worries, I think I actually got the suggestion wrong :).

PR #253, opened 2021-09-14
Add sample-metadata as analysis-runner logging target

illusional:
Implementation:
- Airtable implementation is at risk of 10,000 base limit in not too distant future
- Initial plan to work alongside existing Airtable implementation
- Bad visibility for devs (queryable, but not as easily to flick through as Airtable)

Relies on: https://github.com/populationgenomics/sample-metadata/pull/45


lgruen (approved):
This is surprisingly short! Awesome!

PR #255, opened 2021-09-16
Dataproc: crop cluster ID that can't be longer than 49 chars

vladsaveliev:
Test: https://batch.hail.populationgenomics.org.au/batches/5369


illusional (approved):
Code looks good to me, I don't know the implication of truncating cluster names, could be a recipe for unintentional clashes.

vladsaveliev:
The beginning of the id will be `dataproc-{uuid.uuid4().hex}` in any case, which fully fits within 49 characters, so clashes are super unlickely.

Thanks!

PR #262, opened 2021-09-23
Add environment var

vivbak:


analysis_runner/analysisrunner.py
```diff
@@ -33,6 +34,13 @@ def add_analysis_runner_args(parser=None) -> argparse.ArgumentParser:
         parser = argparse.ArgumentParser('analysis-runner subparser')
 
     add_general_args(parser)
+
+    parser.add_argument(
+        '--environment-variables',
+        required=False,
+        help='A dictionary of environment variables',
+    )
```
illusional, line 42: Repeating args with action=append on an argparse 

```suggestion
    action='append'
    )
```

analysis_runner/analysisrunner.py
```diff
@@ -52,6 +60,7 @@ def run_analysis_runner(
     commit=None,
     repository=None,
     cwd=None,
+    environment_variables=None,
```
illusional, line 63: This would then become:

```suggestion
    environment_variables: List[str]=None,
```

illusional (changes requested):
You've got all the pieces, I think it might be more straight forward on the CLI to be a repeating, equals departed argument, rather than a json dict. Then you don't need to handle a "bad inputs dictionary" case, like a json list, nested dict or int value. eg:

```
-e SM_ENVRONMENT=dev -e OTHERENVKEY=value
```

server/main.py
```diff
@@ -114,6 +115,14 @@ async def index(request):
     job.env('HAIL_BILLING_PROJECT', dataset)
     job.env('DATASET_GCP_PROJECT', dataset_gcp_project)
     job.env('OUTPUT', output_suffix)
+
+    if environment_variables:
+        for env_var_pair in environment_variables:
+            pair = env_var_pair.split('=')
```
illusional, line 121: I think this formatting would be cleaner on the CLI (so keeping your original code here). You might want to consider the case where someone doesn't provide an equals on the CLI, or doesn't provide the map in the correct format.

For example:

```py
if not isinstance(environment_variables, dict):
    raise ValueError('Expected environment_variables to be dictionary')
invalid_env_vars = [f'{k}={v}' for k,v in environment_variables.items() if not isinstance(v, str)]
if len(invalid_env_vars) > 0:
    raise ValueError('Some values of the environment_Variables were not a string, got ' + ', '.join(invalid_env_vars))
```

illusional (changes requested):

analysis_runner/analysisrunner.py
```diff
@@ -111,6 +122,18 @@ def run_analysis_runner(
...
         _cwd = None
 
+    _environment_variables = None
+    if environment_variables:
+        _environment_variables = {}
+        for env_var_pair in environment_variables:
+            try:
+                pair = env_var_pair.split('=')
+                _environment_variables[pair[0]] = pair[1]
+            except IndexError as e:
```
illusional, line 132: This is great to catch!

PR #265, opened 2021-09-28
Add ozrefdb stack

illusional:


lgruen (approved):
Do you also want to delete the previous `refdb` project? (Both through Pulumi `destroy` and afterwards deleting the GCP project.)

PR #269, opened 2021-10-01
Improve error messages of analysis-runner requests

illusional:
Analysis runner is pretty vague on error messages from the client's point of view, and often requires someone to manually check the cloud logs. This PR adds a global exception handler to catch and return meaningful error messages. 

It also adds a _required parameters_ check up front, rather than relying on individual key errors. This change isn't that big, it's mostly un-indenting blocks of code and GH diff tool has not made it look pretty.

Now, the analysis-runner will return a response like the following:

Missing request parameter:
```json
{
    "message": "Missing request parameter(s): cwd, description",
    "success": false
}
```
Failed request to Hail batch:
```json
{
    "message": "unknown billing project fewgenomes__BAD",
    "success": false
}
```
Bad request JSON
```json
{
    "message": "Expecting property name enclosed in double quotes: line 13 column 1 (char 294)",
    "success": false
}
```

### DIFF
--- a/server/cromwell.py
+++ b/server/cromwell.py
@@ -4,7 +4,6 @@ Exports 'add_cromwell_routes', to add the following route to a flask API:
     POST /cromwell: Posts a workflow to a cromwell_url
 """
 import json
-import logging
 import subprocess
 from datetime import datetime
 from shlex import quote
@@ -29,6 +28,17 @@ from util import (
     DRIVER_IMAGE,
     get_cromwell_key,
 )
+
+CROMWELL_REQUIRED_PARAMS = [
+    'accessLevel',
+    'commit',
+    'cwd',
+    'dataset',
+    'description',
+    'output',
+    'repo',
+    'workflow',
+]
 
 
 def add_cromwell_routes(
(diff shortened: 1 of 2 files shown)


server/cromwell.py
```diff
@@ -30,6 +29,17 @@
     get_cromwell_key,
 )
 
+CROMWELL_REQUIRED_PARAMS = [
```
lgruen, line 32: This seems a bit brittle in the sense that we might easily forget to extend this list when adding a parameter.

Previously, we caught the exception that's raised when an element isn't present in the dictionary. I completely agree that the resulting error message wasn't great. However, would there be a way to retrieve the name of the missing key from the exception to create a better error message, making this explicit list redundant?

lgruen (changes requested):

server/cromwell.py
illusional: Yep, it does this as well, I've just moved that error message creation to the generic exception handling bit: 
https://github.com/populationgenomics/analysis-runner/blob/90de4c6493e6c4970439185a67de41085f333c95/server/main.py#L167-L171

I'm happy to remove these required_param checks though, I agree with the brittleness.

illusional: I've removed the required_param checks.

server/cromwell.py
```diff
@@ -94,165 +93,160 @@ async def cromwell(request):  # pylint: disable=too-many-locals
...
+        hail_token = ds_config.get(f'{access_level}Token')
+        service_account_json = get_cromwell_key(
+            dataset=dataset, access_level=access_level
+        )
+        # use the email specified by the service_account_json again
+        service_account_dict = json.loads(service_account_json)
+        service_account_email = service_account_dict.get('client_email')
+        if not service_account_email:
+            raise web.HTTPServerError(
+                reason="The service_account didn't contain an entry for client_email"
```
lgruen, line 117:
```suggestion
                reason='The service_account didn't contain an entry for client_email'
```

server/cromwell.py
lgruen: Cool idea to use a middleware for this!

lgruen (approved):

server/cromwell.py
illusional: This has a single quote in the string “didn’t”, it could be escaped but I think black would resolve it back to double quotes and pylint is happy.

PR #271, opened 2021-10-07
Bump version: Improve cromwell reporting

illusional:
- Fixes `-i` input argument parsing (server is expecting list, received only a string)
- Add pretty colours to Cromwell status reporting
- Fixes <1s time reporting for cromwell job
- Adds `queuedincromwell` status
- Better start / finish time checking for list of calls
- Add a visualise from metadata json file `analysis-runner cromwell visualise <file>` with same config parameters as `status`
- Add `--dry-run` flag to `cromwell submit` which prints a mock curl request.

Todo:
- [x] Fix double reporting in progress view, this is due to grouped calls with a single item in the list

Eg: start of GATK SV:

image


vladsaveliev (approved):
That's so awesome 🙏

PR #272, opened 2021-10-07
Move metadata logging to after batch submit to catch url

illusional:
Useful for linking the instance in the airtable to the actual run:
- New column has already been added to airtable, (eg: https://airtable.com/appOXBjVbfdF4ckHI/tblx9NarwtJwGqTPA/viwIomAHV49Stq5zr/rech6Q1x6ez02MTSy?blocks=hide)
- This works for sample-metadata too.

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -72,25 +72,20 @@ async def index(request):
     # This metadata dictionary gets stored in the metadata bucket, at the output_dir location.
     hail_version = await _get_hail_version()
     timestamp = datetime.datetime.now().astimezone().isoformat()
-    metadata = json.dumps(
-        get_analysis_runner_metadata(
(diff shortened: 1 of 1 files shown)


server/main.py
```diff
@@ -125,6 +120,10 @@ async def index(request):
 
     url = run_batch_job_and_print_url(batch, wait=params.get('wait', False))
 
+    # Publish the metadata to Pub/Sub. Wait for the result before running the batch.
```
lgruen, line 123: The comment is a bit misleading now :)

lgruen (changes requested):

server/main.py
illusional: Thanks for the catch! I've updated the comment :)

PR #283, opened 2021-11-09
Set `HAIL_QUERY_BACKEND=local` on driver image

vladsaveliev:
To make sure query scripts outside dataproc use the spark local backend, and not the incomplete service backend.

cc @KatalinaBobowik


illusional:
I'm okay with this change in the short term because our hail query service isn't fully functional, but when running through the analysis-runner you can use @vivbak's environment variable options (introduced in #262), eg:

```sh
analysis-runner -e HAIL_QUERY_BACKEND=local <other params here>
```

FYI @KatalinaBobowik.

illusional (approved):

vladsaveliev:
Oh, that's nice!

PR #290, opened 2021-11-18
Allow pulumi config to specify additional main upload buckets

illusional:
Allows you to create extra upload buckets with all the same requirements as the regular main-upload bucket carries.

Use for example the following in a pulumi config file:

```yaml
datasets:additional_upload_buckets: '["cpg-another-upload-bucket"]'
```

I created this for something unrelated that's not going ahead, but figured it still might be useful behaviour to add.

### DIFF
--- a/stack/__main__.py
+++ b/stack/__main__.py
@@ -40,7 +40,7 @@ SampleMetadataAccessorMembership = namedtuple(
 )
 
 
-def main():  # pylint: disable=too-many-locals
+def main():  # pylint: disable=too-many-locals,too-many-branches
     """Main entry point."""
 
     # Fetch configuration.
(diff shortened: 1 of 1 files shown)


stack/__main__.py
```diff
@@ -144,21 +144,31 @@ def bucket_member(*args, **kwargs):
...
+        'main-upload': create_bucket(
+            bucket_name('main-upload'), lifecycle_rules=[undelete_rule]
+        )
+    }
+
+    for additional_upload_bucket in (
+        config.get_object('additional_upload_buckets') or ()
+    ):
+        main_upload_buckets[additional_upload_bucket] = create_bucket(
+            additional_upload_bucket, lifecycle_rules=[undelete_rule]
```
vladsaveliev, line 157: Should `additional_upload_bucket` be wrapped in `bucket_name(additional_upload_bucket)`?

vladsaveliev (approved):

stack/__main__.py
illusional: I thought about this, but then you're always restricted to `cpg-{dataset}-...`, which  felt a little limiting, what do you think?

PR #295, opened 2021-11-24
Fix cromwell route

illusional:
- Remove old pubsub remnant, this is now below the workflow submit
- Deploy web server required


vivbak (approved):
Thanks Michael 🙏 !

PR #298, opened 2021-12-08
Fix cromwell submission

illusional:


server/main.py
```diff
@@ -25,6 +25,16 @@
     get_server_config,
 )
 
+logging.basicConfig(level=logging.INFO)
+# do it like this so it's easy to disable
+USE_GCP_LOGGING = True
+if USE_GCP_LOGGING:
+    import google.cloud.logging  # pylint: disable=import-error,no-name-in-module,c-extension-no-member
+
+    client = google.cloud.logging.Client()
```
lgruen, line 34: I'm just curious -- which sort of logs do we get this way that were missing before? Is it the Python exceptions + stack traces?

illusional: I find `logging.*` calls are more consistently in the logs or that they show up quicker - that sounds super vague though. I saw logs from third party apps when it was configured to DEBUG logs which was super useful for actually tracking down where the problem was.

vivbak (approved):
🙏 thanks very much for the fix

PR #299, opened 2021-12-13
Allow humans WRITE access to main resources

illusional:
Per: https://centrepopgen.slack.com/archives/C020PDNH3N0/p1639356527032300?thread_ts=1639355990.032000&cid=C020PDNH3N0


lgruen:
Thanks, I've commented on the Slack thread.

lgruen (changes requested):
Is this obsolete?

illusional:
I think we've had some back and forth and agreed that only machines should really have write access to metadata - It does add some complication when manual intervention is required to modify metadata, but happy to close this.

PR #300, opened 2021-12-14
Automate dataset creation with new_stack script

illusional:
Mostly a bunch of subprocess calls with some minor processing for Hail service account collection.

This script requires you to have created hail service accounts prior to running the script, good future project would be to do that automatically.


stack/new_stack.py
```diff
@@ -0,0 +1,176 @@
...
+        raise ValueError(
+            f'The dataset length must be less than 17 characters (got {len(dataset)})'
+        )
+
+    pulumi_config_fn = f'Pulumi.{dataset}.yaml'
+    if os.path.exists(pulumi_config_fn):
+        raise ValueError('The pulumi config already exists')
+
+    inp = str(
+        input('Please confirm you have created the accounts in Hail Batch (y/n): ')
```
lgruen, line 47: I think `click.confirm` does this?

stack/new_stack.py
```diff
@@ -0,0 +1,176 @@
...
+    ]
+    check_output(command)
+
+
+def create_budget(project_id: str, amount=100):
+    """
+    Create a monthly budget for the project_id
+    """
+    command = [
+        'gcloud',
```
lgruen, line 122: Nice -- that's probably easier to maintain than using the Billing API directly, and doesn't matter performance wise in this case.

stack/new_stack.py
```diff
@@ -0,0 +1,176 @@
...
+    hail_client_emails_by_level = get_client_emails_from_kubectl_output(hail_tokens)
+
+    return hail_client_emails_by_level
+
+
+def get_client_emails_from_kubectl_output(output):
+    """
+    Kubectl gives us 3 json blobs, so use a regex to find the hail emails
+    """
+    responses = RE_CLIENT_EMAIL_MATCHER.findall(output)
```
lgruen, line 166: That's a bit hacky, but parsing the json is kind of painful, I know...

stack/new_stack.py
```diff
@@ -0,0 +1,176 @@
...
+
+    with open(pulumi_config_fn, 'w+', encoding='utf-8') as fp:
+        print(f'Writing to {pulumi_config_fn}')
+        yaml.dump(pulumi_stack, fp, default_flow_style=False)
+
+    command = f"""\
+    export PULUMI_CONFIG_PASSPHRASE=
+    pulumi stack init {dataset}
+    pulumi stack select {dataset}
+    pulumi up
```
lgruen, line 82: Is it worth adding `git` + `gh` commands here as well, to make sure the new config gets merged?

lgruen (approved):
This is great! Could you please also update the documentation on how to use this (and in particular which permissions are required, including how to set up auth for `kubectl`), ideally with a full example that points out not to forget to send a PR for the new stack config?

stack/new_stack.py
```diff
@@ -0,0 +1,176 @@
...
+        *(
+            '--notifications-rule-pubsub-topic',
+            f'projects/{BILLING_PROJECT_ID}/topics/budget-notifications',
+        ),
+    ]
+
+    return check_output(command)
+
+
+def get_hail_service_accounts(dataset: str):
```
lgruen, line 144: Would it be difficult to also create the Hail service accounts and Hail billing projects? Those should just be standard REST endpoints, right?

stack/new_stack.py
illusional: The problem is it returns 3 json blobs, which I couldn't think of a trivial way to work out how to pick up each blob, without seeking and counting braces.

stack/new_stack.py
illusional: I've parameterised this a lot, and it commits automatically (kind of like bumpversion) when you run the step that creates the config.

stack/new_stack.py
illusional: I was hoping so, but I've used the billing-budget API to call this.

stack/new_stack.py
illusional: Changed to click.confirm :)

stack/new_stack.py
lgruen: But that's just because `GET_HAIL_TOKENS` runs a for-loop? Couldn't you instead pull out the loop, so you can parse this more easily?

lgruen:
> The sort of bad caveat to wanting to generate the account JSON is you need to deploy the stack before you generate it, which usually would require a review.

That's probably worth highlighting in a README?

lgruen (changes requested):

stack/new_stack.py
illusional: Yeah that's a much better idea, I've changed it to do this instead

lgruen (approved):
Very nice work! Next it would be good to think about how we can encapsulate the running of this script, so one doesn't need admin permissions in order to invoke this as a data manager.

vladsaveliev (approved):
This is really great!

PR #302, opened 2021-12-15
Improve access_group_cache tolerance to failure

illusional:
There were some errors this morning (my testing indicates too-many-request exceptions) with group member resolution, and this script used exceptions to handle the "user is not a google group" case which caused unexpected downstream errors.

This PR does a few things:

- Allows group member resolution errors to bubble up, and will skip updating a group's secret.
- Catches only the 403 (auth) error during group_parent resolution to indicate a user is not a group
- Improves the group membership checks to identify whether a member is a group (and use that in the previous call)
- Adds an additional simple check for `.iam.gserviceaccount.com` to indicate user is not a group.

### DIFF
--- a/access_group_cache/main.py
+++ b/access_group_cache/main.py
@@ -15,7 +15,14 @@ ANALYSIS_RUNNER_PROJECT_ID = 'analysis-runner'
 app = Flask(__name__)
 
 
-async def _groups_lookup(access_token: str, group_name: str) -> Optional[str]:
+async def _get_group_parent(access_token: str, group_name: str) -> Optional[str]:
+    """
+    Get the group parent (group ID)
+    """
+    if group_name.endswith(".iam.gserviceaccount.com"):
+        # we know it's not a group, because it's a service account
+        return None
+
     async with aiohttp.ClientSession() as session:
         # https://cloud.google.com/identity/docs/reference/rest/v1/groups/lookup
         async with session.get(
--- a/access_group_cache/main.py
+++ b/access_group_cache/main.py
@@ -19,7 +19,7 @@ async def _get_group_parent(access_token: str, group_name: str) -> Optional[str]
     """
     Get the group parent (group ID)
     """
-    if group_name.endswith(".iam.gserviceaccount.com"):
+    if group_name.endswith('.iam.gserviceaccount.com'):
         # we know it's not a group, because it's a service account
         return None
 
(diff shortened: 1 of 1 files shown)


access_group_cache/main.py
```diff
@@ -14,46 +15,70 @@
...
+
     async with aiohttp.ClientSession() as session:
         # https://cloud.google.com/identity/docs/reference/rest/v1/groups/lookup
         async with session.get(
             f'https://cloudidentity.googleapis.com/v1/groups:lookup?'
             f'groupKey.id={urllib.parse.quote(group_name)}',
             headers={'Authorization': f'Bearer {access_token}'},
         ) as resp:
-            if resp.status != 200:
+            if resp.status == 403:
```
lgruen, line 33: Should this ever happen if we already have the filter at the top? Should this instead raise an assertion that contains more info (is there additional status text in `resp` for the 403 case?)?

access_group_cache/main.py
```diff
@@ -14,46 +15,70 @@
...
+    child_groups = []
 
     async with aiohttp.ClientSession() as session:
         page_token = None
         while True:
             # https://cloud.google.com/identity/docs/reference/rest/v1/groups/lookup
             async with session.get(
                 f'https://cloudidentity.googleapis.com/v1/{group_parent}/memberships?'
-                f'pageToken={page_token or ""}',
+                f'view=FULL&pageToken={page_token or ""}',
```
lgruen, line 57: Could you add a comment why we need `view=FULL` here?

access_group_cache/main.py
```diff
@@ -150,6 +189,11 @@ def index():
     )
 
     for group, group_members in zip(groups, all_group_members):
+        if isinstance(group_members, Exception):
+            logging.warning(
```
lgruen, line 193: I think this should be an error. Can we please also set up alerting for this?
(Vivian said it's fairly easy to scan the logs and emit a Slack message when it triggers.)

lgruen (approved):
If this started failing due to "too many requests", we should try to increase the API quota -- the number of requests should really be fairly small. Alternatively, we could try spacing out the requests more, but that's kind of silly as well.

access_group_cache/main.py
illusional: The filter at the top only stops service accounts, users can still be members of the group and they're harder to sanity check. The new 'group' check in the group_membership function should now (I think) stop all these from coming through this channel, but I don't know how to get 100% confidence that it will.

There's no extra information in the response, just `https://cloudidentity.googleapis.com/v1/groups:lookup?groupKey.id=access-group-cache@analysis-runner.iam.gserviceaccount.com`

access_group_cache/main.py
illusional: Yep agreed, I've made it an error and will set-up the log scanner.

PR #309, opened 2022-01-24
Add ohmr4 epilepsy stack

illusional:


lgruen (approved):
Thanks! Just checking whether we also need to add repo permissions for the analysis-runner?

illusional:
We'll need to add it so google-groups permissions are actually propagated correctly. I'll do that in a separate PR due to a configuration change.

PR #318, opened 2022-02-09
Fix dataproc svc name in pulumi

illusional:
@lgruen, I think this is why sometimes the stack failed to deploy intermittently.

### DIFF
--- a/stack/__main__.py
+++ b/stack/__main__.py
@@ -80,7 +80,7 @@ def main():  # pylint: disable=too-many-locals,too-many-branches
     )
 
     # Enable Dataproc until the Hail Query Service is ready.
-    cloudidentity = gcp.projects.Service(
+    _ = gcp.projects.Service(
         'dataproc-service',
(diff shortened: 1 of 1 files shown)


stack/__main__.py
```diff
@@ -80,7 +80,7 @@ def org_role_id(id_suffix: str) -> str:
     )
 
     # Enable Dataproc until the Hail Query Service is ready.
-    cloudidentity = gcp.projects.Service(
+    _ = gcp.projects.Service(
```
lgruen, line 83: Nice catch, but do we actually need the assignment at all?

lgruen (approved):

stack/__main__.py
illusional: The `_ = ` is the python semantic for ignoring the return value. Otherwise some linters might complain you're discarding the result of a function.

lgruen: Right, I think in other places we don't use that pattern though. Fine either way! :)

PR #323, opened 2022-02-14
Add Batch example

vladsaveliev:
Add example of submitting a Batch job with the analysis runner.

cc @hopedisastro 

Test run: https://batch.hail.populationgenomics.org.au/batches/7786/jobs/1


examples/batch/batch.py
```diff
@@ -0,0 +1,66 @@
...
+
+DATASET = os.getenv('DATASET')
+BUCKET = os.getenv('HAIL_BUCKET')
+BILLING_PROJECT = os.getenv('HAIL_BILLING_PROJECT')
+ACCESS_LEVEL = os.getenv('ACCESS_LEVEL')
+
+
+@click.command()
+@click.option('--cram', 'cram_path', 'Input CRAM or BAM file')
+@click.option('--output', 'output_path', 'Output path to write the result')
```
illusional, line 18: Can we build this argument from environment variables, eg:

```python
# or something like this
output_filename = os.path.basename(cram_path) + '-processed.cram'
output_prefix = f'gs://{BUCKET}/{os.getenv("OUTPUT")}/{output_filename}'
```

examples/batch/batch.py
```diff
@@ -0,0 +1,66 @@
...
+            dict=ref_fasta.replace('.fasta', '').replace('.fna', '').replace('.fa', '')
+            + '.dict',
+        )
+    )
+
+    # This image contains basic bioinformatics tools like samtools, bcftools, Picard, etc.
+    j.image('australia-southeast1-docker.pkg.dev/cpg-common/images/bioinformatics:v1-1')
+
+    # For larger CRAMs, request more storage.
+    j.strorage('10G')
```
illusional, line 47:
```suggestion
    j.storage('10G')
```

examples/batch/batch.py
```diff
@@ -0,0 +1,66 @@
...
+    # The command that do the actual job.
+    j.command(
+        f"""
+    samtools view {cram} -T {ref} -L chr21:1-10000 -o -Ocram {j.output_bam}
+    """
+    )
+
+    # Writing the result
+    b.write_output(j.output_bam, output_path)
+    b.run()
```
illusional, line 62:
```suggestion
    # don't wait for the hail batch workflow to complete, otherwise
    # the workflow might get resubmitted if this VM gets preempted.
    b.run(wait=False)
```

examples/batch/batch.py
```diff
@@ -0,0 +1,66 @@
...
+    backend = hb.ServiceBackend(billing_project=BILLING_PROJECT, bucket=BUCKET)
+    b = hb.Batch(backend=backend, default_image=os.getenv('DRIVER_IMAGE'))
+
+    # Adding a job and giving it a descriptive name.
+    j = b.new_job('Subset CRAM')
+
+    # Making sure Hail Batch would localize both CRAM and the correponding CRAI index
+    cram = b.read_input_group(**{'cram': cram_path, 'cram.crai': cram_path + '.crai'})
+
+    # Working with CRAM files requires the refernece fasta
```
illusional, line 32:
```suggestion
    # Working with CRAM files requires the reference fasta
```

illusional: Should we parameterise this reference, or define it as a constant at the top?

examples/batch/batch.py
vladsaveliev: Sure, changed it to use the env variables. Though writing permanent results into the `HAIL_BUCKET` might not be a good idea in practice. But doesn't matter for an example

illusional: Can you tack on `-tmp` on the end of the {BUCKET}?

illusional (approved):

PR #324, opened 2022-02-14
Allow specifying custom container in analysis-runner

illusional:
FYI @vladsaveliev, @lgruen, @MattWellie.

Initial draft, there are a couple of notes:

- I've restricted it to the `cpg-common` artifact registry, we could discuss whitelisting more registries, just a wildcard there is too flexible.
- The container you specify MUST contain gsutil, python3, git
- I've allowed you to specify a number of CPUs and memory, because I imagine people will now perform more types of analysis directly in the _driver_.

EG:

```shell
analysis-runner \
  --dataset fewgenomes \
  --description 'testing container' --output-dir outdir \
  --container australia-southeast1-docker.pkg.dev/cpg-common/images/<image>:latest \
  --cpu 2 --memory '2.2G' \
  python --version
```


analysis_runner/cli_analysisrunner.py
```diff
@@ -35,6 +35,19 @@ def add_analysis_runner_args(parser=None) -> argparse.ArgumentParser:
 
     add_general_args(parser)
 
+    parser.add_argument(
+        '--container',
```
lgruen, line 39: Would `docker_image` be more precise? I.e. doesn't a container usually refer to a specific instance of a Docker image? (Or alternatively just `image`, to mirror the Hail Batch API.)

analysis_runner/cli_analysisrunner.py
```diff
@@ -35,6 +35,19 @@ def add_analysis_runner_args(parser=None) -> argparse.ArgumentParser:
...
+        help=(
+            'Container name, if using standard / full access levels, this must start with '
+            'australia-southeast1-docker.pkg.dev/cpg-common/*.'
+        ),
+    )
+    parser.add_argument('--cpu', type=int, help='Number of CPUs to request')
```
lgruen, line 45: Might also be worth mentioning the Hail Batch docs here, as that explains how to request mCPU, for example. Does that work with Cromwell too, btw?

analysis_runner/cli_analysisrunner.py
```diff
@@ -35,6 +35,19 @@ def add_analysis_runner_args(parser=None) -> argparse.ArgumentParser:
...
+    parser.add_argument(
+        '--memory',
+        help='Amount of memory to request in GB, see hail batch docs for available options. Eg: 4G',
+    )
+
     parser.add_argument(
```
lgruen, line 51: Can we rename this to `env`, to match the Hail Batch API?

server/util.py
```diff
@@ -181,3 +182,13 @@ def write_metadata_to_bucket(
...
+
+
+def validate_container(container: str, is_test: bool):
+    """
+    Check that the image is valid for the access_level
+    """
+    if is_test:
+        return True
+
+    return container.startswith(CONTAINER_IMAGE_PREFIX)
```
lgruen, line 194:
```suggestion
    return is_test or container.startswith(CONTAINER_IMAGE_PREFIX)
```

lgruen (changes requested):

analysis_runner/cli_analysisrunner.py
illusional: Yeah I don't mind image, I don't like `docker` because any OCI compliant image works.

analysis_runner/cli_analysisrunner.py
illusional: Yep, I've allowed you to specify it like hail batch now and added links to docs.
I haven't enabled this for Cromwell, because all it does is prepare and deploy, no processing is done, so nothing should really be able to be configured.

analysis_runner/cli_analysisrunner.py
```diff
@@ -130,13 +155,13 @@ def run_analysis_runner(
...
-    _environment_variables = None
-    if environment_variables:
-        _environment_variables = {}
-        for env_var_pair in environment_variables:
+    _env = None
+    if env:
+        _env = {}
+        for env_var_pair in env:
             try:
                 pair = env_var_pair.split('=')
```
lgruen, line 163:
```suggestion
                pair = env_var_pair.split('=', maxsplit=1)
```

server/main.py
```diff
@@ -54,13 +55,21 @@ async def index(request):
     check_dataset_and_group(server_config, dataset, email)
     repo = params['repo']
     check_allowed_repos(server_config, dataset, repo)
+
+    image = params.get('image') or DRIVER_IMAGE
+    cpus = params.get('cpu', 1)
+    mem = params.get('memory', '1G')
```
lgruen, line 61:
```suggestion
    cpu = params.get('cpu', 1)
    memory = params.get('memory', '1G')
```

server/main.py
```diff
@@ -107,17 +116,20 @@ async def index(request):
...
         dataset=dataset,
         output_suffix=output_suffix,
         metadata_str=json.dumps(metadata),
     )
-    job.image(DRIVER_IMAGE)
+    job.image(image)
+    if cpus:
+        job.cpu(cpus)
+    if mem:
+        job.memory(mem)
```
lgruen, line 131:
```suggestion
    if cpu:
        job.cpu(cpu)
    if memory:
        job.memory(memory)
```

analysis_runner/cli_analysisrunner.py
```diff
@@ -35,9 +35,31 @@ def add_analysis_runner_args(parser=None) -> argparse.ArgumentParser:
 
     add_general_args(parser)
 
+    parser.add_argument(
+        '--image',
+        help=(
+            'Image name, if using standard / full access levels, this must start with '
+            'australia-southeast1-docker.pkg.dev/cpg-common/*.'
```
lgruen, line 42:
```suggestion
            'australia-southeast1-docker.pkg.dev/cpg-common/'
```

lgruen (approved):

PR #325, opened 2022-02-14
An example hail batch workflow

illusional:
Hey @vladsaveliev, I think we committed our test examples at almost the exact same time haha.

I won't change the README yet until yours https://github.com/populationgenomics/analysis-runner/pull/323 is merged, then I'll update the README.

But brief example:

```shell
cd examples/batch
analysis-runner \
  --access-level test \
  --dataset fewgenomes \
  --description "Run Batch" \
  --output-dir "$(whoami)/hello-world" \
  hello.py \
  --name-to-print $(whoami)
```


examples/hailbatch/hello.py
```diff
@@ -0,0 +1,48 @@
+"""
+Test Hail Batch Workflow
+
+
+    cd examples/batch
```
vladsaveliev, line 5: I used a README in my example, but I like the idea of using a docstring more!

vladsaveliev (approved):
Awesome, we got 2 good examples, yours showing how to connect jobs together, mine is an example with bioinfo tools.

PR #326, opened 2022-02-16
Use the cpg-hail pip package

vladsaveliev:
Adjusting Dockefiles to install the `cpg-hail` pip package instead of the conda package.

We can then further modify CI to build analysis-runner for pip as well

### DIFF
--- a/dataproc/Dockerfile
+++ b/dataproc/Dockerfile
@@ -6,4 +6,4 @@ RUN apt update && \
     curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - && \
     apt update && apt install -y google-cloud-sdk && \
     apt clean && rm -rf /var/lib/apt/lists/* && \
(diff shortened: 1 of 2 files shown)


dataproc/Dockerfile
```diff
@@ -6,4 +6,4 @@ RUN apt update && \
     curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - && \
     apt update && apt install -y google-cloud-sdk && \
     apt clean && rm -rf /var/lib/apt/lists/* && \
-    pip3 install hail==0.2.73
+    pip3 install cpg-hail
```
illusional, line 9: Should we pin the version?

illusional (approved):

dataproc/Dockerfile
vladsaveliev: Don't think it's really needed for the dataproc helper, it's fine to take the latest

PR #334, opened 2022-02-22
Upgrade from Python 3.9 to 3.10

lgruen:
Now working thanks to your `tabulate` fix the other day 🙇

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -4,7 +4,7 @@ channels:
   - bioconda
   - conda-forge
 dependencies:
-  - python=3.9
+  - python=3.10
   - pip
   - cpg-utils  # Not yet in pypi.
   - pip:


vladsaveliev:
Thanks @lgruen! 

However, we actually just dropped the `environment-dev.yml` when switching to PyPI - that PR was merged just after you submitted this one. Now you can setup the full environment with `pip install -e requirements-dev.txt` and `pip install analysis-runner` or `pip install .`, so no need to provide a `environment-dev.yml` for conda. The package should work with py3.10

lgruen:
@vladsaveliev That's great, thanks! Do I understand correctly that `pip install .` will install the dependencies listed at `https://github.com/populationgenomics/analysis-runner/blob/main/setup.py#L31`?

vladsaveliev:
@lgruen, that's right! It installs everything in `install_requires`.

PR #336, opened 2022-02-28
pip likes double equals for version pins

illusional:
See L1686 for failure: https://github.com/populationgenomics/analysis-runner/runs/5353912210?check_suite_focus=true


vladsaveliev (approved):
Thanks!!

PR #341, opened 2022-03-08
Install Google Cloud SDK in the driver image

lgruen:
Fixes the currently broken build: https://github.com/populationgenomics/analysis-runner/runs/5460052892?check_suite_focus=true

### DIFF
--- a/driver/Dockerfile
+++ b/driver/Dockerfile
@@ -30,6 +30,4 @@ RUN apt-get update && apt-get install -y apt-transport-https bash bzip2 ca-certi
     cpg-hail==$HAIL_VERSION \
     cpg-utils==2.1.1 \
     analysis-runner \
-    sample-metadata==4.4.0 && \
-    # hailctl dataproc uses gcloud beta dataproc.
-    gcloud -q components install beta
+    sample-metadata==4.4.0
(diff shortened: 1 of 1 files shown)


driver/Dockerfile
```diff
@@ -6,28 +6,30 @@ ENV MAMBA_ROOT_PREFIX /root/micromamba
...
-        phantomjs \
-        r-argparser \
-        r-base=4.1.1 \
-        r-essentials \
-        r-tidyverse \
-        selenium \
-        skopeo \
-        statsmodels && \
+    -c cpg -c bioconda -c conda-forge \
+    bokeh \
```
illusional, line 19: Is there a reason for removing this formatting? I think it's cleaner as it shows it's arguments for the micromamba install.

illusional (approved):
The change looks fine to me, do we know why it started spontaneously failing?
I left a comment about removing the indentation, my personal preference is to keep the indentation.

lgruen:
> The change looks fine to me, do we know why it started spontaneously failing? I left a comment about removing the indentation, my personal preference is to keep the indentation.

I don't think it's spontaneous -- this has been broken for a while, probably in the context of the recent conda --> pip changes: https://github.com/populationgenomics/analysis-runner/actions/workflows/hail_update.yaml
CleanShot 2022-03-08 at 17 14 22@2x

driver/Dockerfile
lgruen: That's just what VS Code does auto-formats Dockerfiles for me... Don't have a strong opinion myself.

lgruen: Let me try https://github.com/microsoft/vscode-docker/pull/2976

PR #344, opened 2022-03-09
Use a slimmer default driver image that doesn't contain conda / R

lgruen:


illusional (commented):
The change by itself looks good to me, but I have a few thoughts:

- I think finding and specifying the new R driver image is going to be painful, where for the most part analysts don't care, they just want the latest version with latest base packages.
- How many DRIVER-like images are we foreseeing, could we make it extensible so there's a list of names / extensions, they all get build automatically and taught to the server by env vars, and then you could specify something like `--image DRIVER:R` / `--image DRIVER:myextension`, and the server replaces them with the most up-to-date tag?

lgruen:
> The change by itself looks good to me, but I have a few thoughts:
> 
> * I think finding and specifying the new R driver image is going to be painful, where for the most part analysts don't care, they just want the latest version with latest base packages.
> * How many DRIVER-like images are we foreseeing, could we make it extensible so there's a list of names / extensions, they all get build automatically and taught to the server by env vars, and then you could specify something like `--image DRIVER:R` / `--image DRIVER:myextension`, and the server replaces them with the most up-to-date tag?

I hope that we don't have many different driver variants within this repo and that the R driver won't be updated frequently. So maybe the easiest solution would be to manually tag the R driver image on the rare occasion that it does get changed? E.g. I could tag `australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:53e387bb282c115f908963d41acd7e0369d46456-hail-0.2.85.post2` manually as `australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.0.0` for now.

The main other use case for the custom driver images I'd foresee are pipelines (like Matt's reanalysis pipeline), for which the image would be built outside of this repo's CI anyhow.

lgruen:
> > The change by itself looks good to me, but I have a few thoughts:
> > 
> > * I think finding and specifying the new R driver image is going to be painful, where for the most part analysts don't care, they just want the latest version with latest base packages.
> > * How many DRIVER-like images are we foreseeing, could we make it extensible so there's a list of names / extensions, they all get build automatically and taught to the server by env vars, and then you could specify something like `--image DRIVER:R` / `--image DRIVER:myextension`, and the server replaces them with the most up-to-date tag?
> 
> I hope that we don't have many different driver variants within this repo and that the R driver won't be updated frequently. So maybe the easiest solution would be to manually tag the R driver image on the rare occasion that it does get changed? E.g. I could tag `australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:53e387bb282c115f908963d41acd7e0369d46456-hail-0.2.85.post2` manually as `australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.0.0` for now.
> 
> The main other use case for the custom driver images I'd foresee are pipelines (like Matt's reanalysis pipeline), for which the image would be built outside of this repo's CI anyhow.

Based on our discussion, I've now derived the R image directly from `base` and only build / tag it manually. We should revisit this if we add sample metadata API bindings for R.

illusional (approved):
This will save a whole bunch of layers every time we update the driver Dockerfile!

PR #349, opened 2022-03-14
Bump version: 2.16.0 → 2.17.0: Set HAIL_SHA and HAIL_JAR_URL to enable the Query Service

lgruen:
Depends on SHA input being set through https://github.com/populationgenomics/hail/pull/149.

Context: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1646979994175289?thread_ts=1646887215.452969&cid=C018KFBCR1C


.github/workflows/hail_update.yaml
```diff
@@ -8,6 +8,9 @@ on:
       hail_version:
         description: 'Hail version (as uploaded to PyPI)'
         required: true
+      hail_sha:
```
illusional, line 11: Adding more inputs makes this more difficult to manually release, eg: if we manually release after updating SM. It also makes it impractical to automatically trigger a release from SM repo, because it would need to find the input for two hail variables.

The hail team have a deploy script which automatically applies a change and triggers a pull request, which might be one way to do this:

https://github.com/hail-is/hail/blob/660616e51f76ca244c2d4a6081ed169d740c10cc/hail/scripts/deploy.sh#L141-L148

.github/workflows/hail_update.yaml
```diff
@@ -17,8 +20,8 @@ jobs:
       DOCKER_BUILDKIT: 1
       BUILDKIT_PROGRESS: plain
       CLOUDSDK_CORE_DISABLE_PROMPTS: 1
-      DRIVER_IMAGE: australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:${{ github.sha }}-hail-${{ github.event.inputs.hail_version }}
-      SERVER_IMAGE: australia-southeast1-docker.pkg.dev/analysis-runner/images/server:${{ github.sha }}-hail-${{ github.event.inputs.hail_version }}
+      DRIVER_IMAGE: australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:${{ github.sha }}-hail-${{ github.event.inputs.hail_version }}-${{ github.event.inputs.hail_sha }}
```
illusional, line 23: Could we make these in the version the short sha? Docker tags have a max 128 length sha, and 2 shas + other stuff is sitting around 95.

.github/workflows/hail_update.yaml
```diff
@@ -52,7 +55,7 @@ jobs:
 
     - name: "deploy server-test"
       run: |
-        gcloud run deploy server-test --region australia-southeast1 --no-allow-unauthenticated --platform managed --set-env-vars=DRIVER_IMAGE=$DRIVER_IMAGE --image $SERVER_IMAGE
+        gcloud run deploy server-test --region australia-southeast1 --no-allow-unauthenticated --platform managed --set-env-vars=DRIVER_IMAGE=$DRIVER_IMAGE,TZ=Australia/Sydney,HAIL_SHA=${{ github.event.inputs.hail_sha }} --image $SERVER_IMAGE
```
illusional, line 58: TZ is new, was there something else that meant we had to add that?

illusional (approved):

.github/workflows/hail_update.yaml
lgruen: This is just for consistency with the prod version.

.github/workflows/hail_update.yaml
lgruen: Hmm, then we'd have to extract the short SHA from the full SHA using an additional step that extracts those bytes. Certainly doable, but alternatively I could also strip the `hail_version`, as that's redundant given `hail_sha`.

.github/workflows/hail_update.yaml
lgruen: Wouldn't that be equivalent to a dummy commit to our Hail repo, as that would also trigger https://github.com/populationgenomics/hail/blob/main/.github/workflows/package.yaml, which in turn runs the analysis-runner workflow with the right inputs?

lgruen: As discussed over Zoom, we could try to replace this with a mechanism that gets the HEAD SHA + version by checking out the Hail repository. However, now that @vladsaveliev suggested that we might actually want to use the standard `hail` package in PyPI instead of `cpg-hail`, I'm not so sure anymore...

.github/workflows/hail_update.yaml
lgruen: I don't think `TZ` is used anywhere, so I'll remove this in a follow-up PR.

Issue #329, opened 2022-02-20
Better support external collaborators (read-only depends_on)


illusional:
- Create new readonly depends_on dependency
- Also ensure this catches the sample-metadata read permissions
- At the same time, allow non populationgenomics repositories
- Still enforce _commit must be on main to use standard / full_.

illusional:
Addressed in: https://github.com/populationgenomics/cpg-infrastructure/pull/33

PR #354, opened 2022-03-16
Allow depends_on relationship to work with sample-metadata

illusional:
- Allow sample-metadata-{main,test}-{read,write} groups to work transitively
- Adds export for sm groups (to allow)
- Dataset with depends on adds itself to relevant depends on sm group


lgruen (changes requested):
Given https://github.com/populationgenomics/analysis-runner/issues/329, won't we need to change our setup to explicit read / write groups anyhow? I'm asking because that might present an opportunity to remove the SM-specific groups entirely, right?

illusional:
Closing as this was addressed in: https://github.com/populationgenomics/cpg-infrastructure/pull/33

PR #358, opened 2022-03-18
Quote dataproc --packages opt

vladsaveliev:
To support `>` and `[]` syntax in package specs; otherwise it silently starts a cluster without packages, like here: https://batch.hail.populationgenomics.org.au/batches/17781/jobs/1

### DIFF
--- a/analysis_runner/dataproc.py
+++ b/analysis_runner/dataproc.py
@@ -197,7 +197,7 @@ def _add_start_job(  # pylint: disable=too-many-arguments
             f'--secondary-worker-boot-disk-size={secondary_worker_boot_disk_size}'
         )
     if packages:
-        start_job_command.append(f'--packages={",".join(packages)}')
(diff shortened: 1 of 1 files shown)


analysis_runner/dataproc.py
```diff
@@ -197,7 +197,7 @@ def _add_start_job(  # pylint: disable=too-many-arguments
             f'--secondary-worker-boot-disk-size={secondary_worker_boot_disk_size}'
         )
     if packages:
-        start_job_command.append(f'--packages={",".join(packages)}')
+        start_job_command.append(f'--packages=\'{",".join(packages)}\'')
```
lgruen, line 200: Should we use `shlex.quote` instead?

lgruen (changes requested):

lgruen:
Don't forget to bump the patch version.

lgruen (approved):

PR #372, opened 2022-03-23
Dataproc: use the populationgenomics wheel built for ES 8.x.x

vladsaveliev:


analysis_runner/dataproc.py
```diff
@@ -17,9 +17,16 @@
...
 
+HAIL_VERSION = '0.2.91'
 DATAPROC_IMAGE = (
-    'australia-southeast1-docker.pkg.dev/analysis-runner/images/dataproc:hail-0.2.91'
+    f'australia-southeast1-docker.pkg.dev/analysis-runner/images/'
+    f'dataproc:hail-{HAIL_VERSION}'
 )
+# Wheel built from https://github.com/populationgenomics/hail
+# The difference from the official build is the version of ElasticSearch:
+# We use 8.x.x, and Hail is built for 7.x.x by default.
```
lgruen, line 27: We're planning to upstream that though, right?

lgruen (approved):
Thank you!

analysis_runner/dataproc.py
vladsaveliev: I asked that in zulip, they don't want to change the ElasticSearch version because it's bound to what is used in the Broad:
https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/export_elasticsearch.3A.20.22Remove.20type.20name.20to.20continue.22/near/276321465

PR #377, opened 2022-03-25
Allow email access list overrides for web buckets

lgruen:
If someone isn't in the generic access list for the web bucket, they can be listed in a `.access` file in the first subdirectory instead. This is helpful to give external collaborators access to particular files.

Context: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1648103614839679?thread_ts=1647580532.711369&cid=C018KFBCR1C

### DIFF
--- a/web/main.py
+++ b/web/main.py
@@ -45,6 +45,8 @@ def handler(dataset=None, filename=None):
             audience=IAP_EXPECTED_AUDIENCE,
             certs_url='https://www.gstatic.com/iap/verify/public_key',
         )
+        # Use allAuthenticatedUsers for the IAP configuration to make this
+        # work for arbitrary users.
         email = decoded_jwt['email']
     except Exception:  # pylint: disable=broad-except
         logger.exception('Failed to extract email from ID token')
--- a/web/main.py
+++ b/web/main.py
@@ -71,19 +71,21 @@ def handler(dataset=None, filename=None):
     ).split(',')
 
     if email not in members:
-        # Second chance: if there's a '.access' file in the file's directory,
+        # Second chance: if there's a '.access' file in the first subdirectory,
         # check if the email is listed there.
-        access_filename = os.path.join(os.path.dirname(filename), '.access')
-        blob = bucket.get_blob(access_filename)
-        if blob is None:
-            logger.warning(f'{email} is not a member of the {dataset} access group')
-            abort(403)
-        access_list = blob.download_as_string().splitlines()
-        if email not in access_list:
-            logger.warning(
-                f'{email} is not in {dataset} access group or {access_filename}'
-            )
-            abort(403)
+        split_subdir = filename.split('/', maxsplit=1)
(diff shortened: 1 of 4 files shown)


web/main.py
```diff
@@ -43,17 +62,32 @@ def handler(dataset=None, filename=None):
...
     members = read_secret(
         dataset_project_id, f'{dataset}-web-access-members-cache'
     ).split(',')
+
     if email not in members:
-        logger.warning(f'{email} is not a member of the {dataset} access group')
-        abort(403)
+        # Second chance: if there's a '.access' file in the first subdirectory,
+        # check if the email is listed there.
+        split_subdir = filename.split('/', maxsplit=1)
```
illusional, line 76: I think using os.path.dirname is cleaner for the downstream code
```suggestion
        subdir = os.path.dirname(filename)
```

MattWellie (approved):

web/main.py
```diff
@@ -43,17 +62,32 @@ def handler(dataset=None, filename=None):
...
+        if len(split_subdir) == 2 and split_subdir[0]:
+            access_list_filename = f'{split_subdir[0]}/.access'
```
illusional, line 78:
```suggestion
            access_list_filename = os.path.join(subdir, '.access')
```

web/main.py
```diff
@@ -43,17 +62,32 @@ def handler(dataset=None, filename=None):
...
+            blob = bucket.get_blob(access_list_filename)
+            if blob is None:
+                logger.warning(f'{email} is not a member of the {dataset} access group')
+                abort(403)
+            access_list = blob.download_as_string().decode('utf-8').splitlines()
```
illusional, line 83: Does `download_as_string` download it as binary, you've got a decode here which seems strange

illusional (approved):
Kind of awkward to implement, but I think a `.access` file only giving you access to the current dir is going to get confusing.

web/main.py
lgruen: Yes, indeed. This has been fixed with newer versions of the library, but unfortunately `hail` (which `cpg-utils` depends on) pins an old version of `google-cloud-storage`...

web/main.py
lgruen: I only want the first subdirectory down from the top-level though, not the `.access` file further down. I thought about allowing those files at any subdirectory level, but then you'd have to walk up the tree and check for blob existence on every level.

web/main.py
lgruen: At the moment, `split_subdir[0]` can't contain any slashes, so `os.path.join` doesn't really help. Would be cleaner if we used `dirname` as you suggested, but that would have different semantics.

web/main.py
MattWellie: This was a fun one to find out 🤦 the d_as_string downloads bytes, and the d_as_bytes doesn't exist 🎉

PR #380, opened 2022-03-29
No need to set HAIL_JAR_URL anymore

lgruen:
See https://github.com/populationgenomics/hail/commit/f242c348c8503310352be2ab6a14f356db32467c, which automates this.

`cpg-utils` will be updated in a separate PR.


vladsaveliev (approved):

lgruen:
We'll need to revert this, now that the Hail upstream merge has been reverted: https://github.com/populationgenomics/hail/pull/177

PR #392, opened 2022-04-19
Group cache v2

illusional:
- [ ] Devise new groups and permissions to satisfy:
    - [x] Removing sample-metadata specific groups
    - [x] Support read-only depends on #329
    - [ ] Get it approved
- [ ] Implement new groups and permissions
- [x] Refactor group cacher to use GCS instead of secrets
    - [ ] Add a server that performs this check
        - [ ] Implement better support for this in cpg-utils: https://github.com/populationgenomics/cpg-utils/pull/29
    - [ ] Refactor other pieces of code:
        - [x] web-accessor
        - [x] sample-metadata: https://github.com/populationgenomics/sample-metadata/pull/185
        - [x] analysis-runner server
    - [x] Write migration from secret storage
- [x] Rewrite access-group-cache to be more efficient

Initial results for rewrite of access-group-cache is promising:

```
BEFORE: Looked up in 25.1s (2314 API requests)
AFTER: Looked up in 2 rounds in 6.4s (491 API requests)
```


illusional:
Closed, addressed in a few different ways that culminated in: https://github.com/populationgenomics/cpg-infrastructure/pull/33

Allowed for pulumi to know the full set of group members at deploy time. No need for a poll function, doesn't use secrets and added code into cpg-utils for getting group members

PR #393, opened 2022-04-21
Fix serialising cromwell watch_workflow: import subprocess

vladsaveliev:
Importing `subprocess` inside `watch_workflow`, otherwise dill fails to serialise `watch_workflow`. For the same reason, re-defining `get_cromwell_oauth_token` inside `watch_workflow`, because it uses `subprocess` as well.


illusional (approved):

vivbak:
❤️ 🙏

PR #396, opened 2022-04-22
Cromwell: fix running watch_workflow from job

vladsaveliev:
Giving up fixing pickling, switching to a bash job to run watch_workflow. That's probably a bad way to do that, but I really want to unblock ourselves.

Test run: https://batch.hail.populationgenomics.org.au/batches/71003/jobs/2


analysis_runner/cromwell.py
```diff
@@ -629,15 +664,3 @@ def _copy_resource_group_into_batch(
         )
 
     return output_filename
-
-
-def get_cromwell_oauth_token():
```
vivbak, line 634: Would removing this entirely cause an issue in server/cromwell.py? 
https://github.com/populationgenomics/analysis-runner/runs/6398752286?check_suite_focus=true#step:5:43

vivbak (commented):
Thanks so much Vlad 🙏 🙏 🙏  Just one question from me.

analysis_runner/cromwell.py
vladsaveliev: Good catch!

vladsaveliev: Thanks :) Fixed

vivbak (approved):
🙏 🙏 🙏

illusional (approved):
Sucks that it came to this :((

PR #400, opened 2022-05-09
Bump version: 2.26.4 → 2.26.5: Add REFERENCE_PREFIX

vladsaveliev:
Similar to CPG_IMAGE_REGISTRY_PREFIX/IMAGE_REGISTRY_PREFIX, add CPG_REFERENCE_PREFIX/REFERENCE_PREFIX. Goes along with `reference_path` added here: https://github.com/populationgenomics/cpg-utils/pull/32


MattWellie (approved):
LGTM

one note - if other sites may use this cpg-utils package, should we keep the CPG- prefix on these env variables? (I know the package is called CPG-utils, so I think the answer is yes)

equally, a CPG deployment environment would only expect one 'driver image', so that would be an equally good ENV variable to set, which could make this a more versatile package

PR #408, opened 2022-05-11
Set the correct encryption salt before committing a new stack file to git

lgruen:
Note that this is a stacked PR.


illusional (approved):
I feel like this could cause merge conflicts, as long as the import statement (not present in this PR) is still there, all good from my side.

PR #409, opened 2022-05-18
Update python version to 3.10

vivbak:


illusional (approved):
LGTM, can you post on team-data to tell people about this change - hopefully this has no python version problems :)

PR #410, opened 2022-05-18
Write config blob instead of setting many environment variables

lgruen:
Depends on `cpg-utils.config`, which gets added in https://github.com/populationgenomics/cpg-utils/pull/35.

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -4,28 +4,27 @@ import datetime
 import json
 import logging
 from shlex import quote
-
 import hailtop.batch as hb
 from aiohttp import web
-
 from analysis_runner.git import prepare_git_job
 from cromwell import add_cromwell_routes
 from util import (
     DRIVER_IMAGE,
     IMAGE_REGISTRY_PREFIX,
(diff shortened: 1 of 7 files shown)


server/main.py
```diff
@@ -133,17 +132,27 @@ async def index(request):
...
+            'bucket': hail_bucket,
+        },
+        'workflow': {
+            'access_level': access_level,
+            'dataset': dataset,
+            'dataset_gcp_project': dataset_gcp_project,
+            'driver_image': DRIVER_IMAGE,
+            'image_registry_prefix': IMAGE_REGISTRY_PREFIX,
+            'reference_prefix': REFERENCE_PREFIX,
+            'output_prefix': output_prefix,
```
vladsaveliev, line 148: `CPG_WEB_URL_TEMPLATE` added here https://github.com/populationgenomics/analysis-runner/pull/411/files is missing, might wanna merge from main :)

vladsaveliev (approved):

PR #422, opened 2022-05-30
Bump version: 2.27.3 → 2.28.0: Write config path to metadata entry

lgruen:


illusional (approved):
Change looks good, but why not write the config directly into the metadata as an entry in the dictionary?

lgruen:
> Change looks good, but why not write the config directly into the metadata as an entry in the dictionary?

I guess we could, but that might grow to hundreds of parameters for some pipelines. @vladsaveliev and I were chatting about merging pipeline-specific options with this structure, so this could grow and bloat the metadata DB otherwise.

PR #424, opened 2022-05-30
Bump version: 2.29.0 → 2.30.0: Update the config specified in the request

lgruen:
@vladsaveliev FYI


server/util.py
```diff
@@ -210,3 +211,15 @@ def write_config(config: MutableMapping[str, Any]) -> str:
     with config_path.open('w') as f:
         toml.dump(config, f)
     return str(config_path)
+
+
+def update_dict(d1: MutableMapping, d2: MutableMapping) -> None:
+    """Updates the d1 dict with the values from the d2 dict recursively in-place."""
+    for k, v2 in d2.items():
+        v1 = d1.get(k)
+        if isinstance(v1, collections.abc.MutableMapping) and isinstance(
```
vladsaveliev, line 220: Code style wise, I personally prefer type-hinting dicts as `typing.Dict` (or even better with `dict` in python 3.10, and avoid importing anything for hints altogether)

vladsaveliev: I know you mentioned `toml.load` returns a `collections.abc.MutableMapping`, but I think calling `dict` to construct standard dict from it should work for mypy

vladsaveliev (approved):
Thank you!

server/util.py
lgruen: Thanks, this is much cleaner! I didn't know that Python 3.10 lets you use `dict` that way as a type hint, too.

Do you happen to know whether `dict[str]` is equivalent to `Dict[str, Any]`? I couldn't find an explanation in the docs.

vladsaveliev: `dict[str]` wouldn't even work, it's ambiguous if `str` refers to keys or to values. That's what `mypy` throws:
```
error: "dict" expects 2 type arguments, but 1 given
```

vladsaveliev: And more on using built-in types for hints: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#built-in-types

PR #426, opened 2022-05-31
Fix config access when none is given

lgruen:


vladsaveliev (approved):
That's really odd! Did `param` somehow already contain `config` value of non-dict type?

lgruen:
> That's really odd! Did `param` somehow already contain `config` value of non-dict type?

It's because we set the config default value to `None` in the CLI here: https://github.com/populationgenomics/analysis-runner/blob/main/analysis_runner/cli_analysisrunner.py#L184

So it's actually `None` in the JSON parameters, which is different from not being present in the dictionary ;).

PR #428, opened 2022-06-02
Use `update_dict` from cpg-utils

vladsaveliev:
Moved the function into cpg-utils as it's reused in cpg-utils and cpg-pipes: https://github.com/populationgenomics/cpg-utils/pull/46


lgruen (approved):

vladsaveliev:
@lgruen please merge when you don't have comments

PR #431, opened 2022-06-07
Allow extra SAs to access sample-metadata

illusional:
Addresses internal request: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1654564802406569

SAs don't need access to IAP, because sample-metadata package uses the cloud run link directly.

### DIFF
--- a/stack/__main__.py
+++ b/stack/__main__.py
@@ -1127,13 +1127,23 @@ def main():  # pylint: disable=too-many-locals,too-many-branches
 
 def _get_name_from_external_sa(email: str, suffix='.iam.gserviceaccount.com'):
     """
-    Convert service account email to name + some filterin
+    Convert service account email to name + some filtering.
 
+    >>> _get_name_from_external_sa('my-service-account@project.iam.gserviceaccount.com')
+    'my-service-account-project'
+
+    >>> _get_name_from_external_sa('yourname@populationgenomics.org.au')
+    'yourname'
+
+    >>> _get_name_from_external_sa('my.service-account+extra@domain.com')
+    'my-service-account-extra'
     """
-    if not email.endswith(suffix):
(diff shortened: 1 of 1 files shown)


stack/__main__.py
```diff
@@ -41,6 +51,8 @@
     ['name', 'member_key', 'permissions'],
 )
 
+NON_NAME_REGEX = re.compile(r'[^A-z0-9_-]')
```
lgruen, line 54:
```suggestion
NON_NAME_REGEX = re.compile(r'[^A-Za-z0-9_-]')
```

stack/__main__.py
```diff
@@ -1081,5 +1125,16 @@ def find_service_account(kind: str, access_level: str) -> Optional[str]:
         )
 
 
+def _get_name_from_external_sa(email: str, suffix='.iam.gserviceaccount.com'):
+    """
+    Convert service account email to name + some filterin
+
+    """
```
lgruen, line 1132:
```suggestion
    """Convert service account email to name + some filtering."""
```

lgruen (approved):

stack/__main__.py
illusional: I'm not following this one, doesn't `A-z` cover `A-Za-z`?

stack/__main__.py
illusional: Thank you! I added whitespace for doctest, but forgot to add it (added now) :)

stack/__main__.py
lgruen: Definitely, if you consider a particular ordering. However, that's not how regex character classes are typically defined: https://github.com/google/re2/wiki/Syntax

illusional: Ooh, learn something new every day!

PR #436, opened 2022-06-15
Fix: import `update_dict` from cpg-utils

vladsaveliev:


MattWellie (approved):

illusional (approved):
Looks good, but just waiting for tests to pass

PR #439, opened 2022-06-23
Install pyarrow

illusional:
Pyarrow is required for hail's to_parquet function - so driver is well placed to install it IMO.

### DIFF
--- a/driver/Dockerfile.hail
+++ b/driver/Dockerfile.hail
@@ -34,6 +34,7 @@ RUN apt update && \
         sample-metadata \
         selenium \
         statsmodels && \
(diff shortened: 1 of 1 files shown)


KatalinaBobowik (approved):
👍

driver/Dockerfile.hail
```diff
@@ -34,6 +34,7 @@ RUN apt update && \
         sample-metadata \
         selenium \
         statsmodels && \
```
lgruen, line 36: There's an extra `&&` here.

Could we keep this list sorted, please?

PR #453, opened 2022-08-01
Fix TZ=Australia backend

MattWellie:
Hail's backend determination requires a full String match to service, batch, local, etc. 
This line caused the backend env variable to be `service   TZ="Australia"` so a full match was not possible

context: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1659395182183599

### DIFF
--- a/driver/Dockerfile.hail
+++ b/driver/Dockerfile.hail
@@ -4,7 +4,7 @@ FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.2
 ARG HAIL_SHA
 
 ENV HAIL_QUERY_BACKEND service
-ENV TZ="Australia"
+ENV TZ Australia
 
 RUN apt update && \
     DEBIAN_FRONTEND=noninteractive apt install -y \
--- a/driver/Dockerfile.hail
+++ b/driver/Dockerfile.hail
@@ -3,8 +3,8 @@ FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.2
 
 ARG HAIL_SHA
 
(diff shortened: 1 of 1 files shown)


driver/Dockerfile.hail
```diff
@@ -3,8 +3,8 @@ FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.2
 
 ARG HAIL_SHA
 
-ENV HAIL_QUERY_BACKEND service \
-    TZ=Australia
+ENV HAIL_QUERY_BACKEND service
+ENV TZ="Australia"
```
vladsavelyev, line 7: I guess both ways work, but I'd use the same format for consistency:
```suggestion
ENV TZ Australia
```

vladsavelyev: (also following the format of these statements https://github.com/populationgenomics/analysis-runner/blob/3c4f1e4ac95ec2f33503d81677bec1144553746a/driver/Dockerfile.r)

vladsavelyev (approved):

driver/Dockerfile.hail
```diff
@@ -3,8 +3,8 @@ FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.2
 
 ARG HAIL_SHA
 
-ENV HAIL_QUERY_BACKEND service \
-    TZ=Australia
+ENV HAIL_QUERY_BACKEND service
+ENV TZ Australia
```
lgruen, line 7: That's not even a valid timezone though, right? I suspect this was just a copy and paste issue or maybe was needed before `DEBIAN_FRONTEND=noninteractive` was added. Can we just remove this environment variable completely?

PR #455, opened 2022-08-04
Dataproc: update Hail version pin to 0.2.95

vladsavelyev:


lgruen (approved):

lgruen:
Note that there's been a regression in 0.2.96 for which the fix hasn't been released yet in 0.2.97: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1657511417463239?thread_ts=1657173965.254659&cid=C030X7WGFCL

vladsavelyev:
Oh right, I actually hit this error, but assumed that `densify` didn't work just because my matrix table was split and didn't have LGT firleds. Okay, let's hold until the next version is released.

vladsavelyev:
@lgruen, do you think it's okay to pin to 0.2.95 (while we are waiting for a new version)? It works for me fine.

It also seems that the code that compresses pyfiles is redundant after https://github.com/hail-is/hail/pull/10863

lgruen:
> @lgruen, do you think it's okay to pin to 0.2.95 (while we are waiting for a new version)? It works for me fine.

Sure, sounds good!

PR #458, opened 2022-08-10
Add cellregmap pipeline

illusional:


tokens/repository-map.json
```diff
@@ -152,7 +152,8 @@
         "tob-wgs",
         "sv-workflows",
         "production-pipelines",
-        "sample-metadata"
+        "sample-metadata",
+        "CellRegMap_pipeline"
```
lgruen, line 156: Can we rename this repository for naming consistency with the other repositories? E.g. `cell-reg-map-pipeline` or so, or even `CellRegMap-pipeline`, in case `CellRegMap` is an existing name.

lgruen (approved):

tokens/repository-map.json
illusional: Thanks! I've renamed to cellregmap-pipeline :)

PR #459, opened 2022-08-10
Add udn-aus stack

illusional:
Created with new_stack script


lgruen (approved):
Do you also want to add a dependency of seqr on this stack?

illusional:
Yes. very much, thanks @lgruen!

PR #463, opened 2022-08-21
Add udn-aus-trainin stack

illusional:


lgruen (approved):
Does seqr need to have a dependency on this?

illusional:
Yes, this dataset is joint-called with other seqr datasets, and loaded into our production instance of seqr for training.

PR #464, opened 2022-08-22
Enable release for validation stack

illusional:
Context: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1661124751453449


MattWellie (approved):
LGTM

MattWellie:
@illusional will this just add the extra bucket, or recreate the whole validation stack from scratch?

illusional:
Pulumi just makes changes, so it's only added the `gs://cpg-validation-release-requester-pays` bucket (with a few permissions). You'll have to use analysis-runner to move files to here.

lgruen:
Note that `release` is a **public** bucket, IIRC. Is that really what we want?

lgruen:
Just checked: it doesn't default to being public, so that should be okay. However, might be worth taking into account for your permissions revamp, @illusional -- this should probably be configurable, too.

PR #465, opened 2022-08-22
Pin selenium version

lgruen:
To avoid `AttributeError: module 'selenium.webdriver' has no attribute 'PhantomJS'`.


KatalinaBobowik (approved):
👍

PR #473, opened 2022-08-24
Add shared project (for data egress protection)

illusional:
Configurable on the dataset's config:

```yaml
  datasets:enable_egress_project: "true"
  datasets:egress_budget: 123
```

- Creates a new project with format `cpg-{dataset}-egress` to avoid any name clashes
- Creates a budget for that project
- Creates a service account in that new project
- Allows that service-account to access requester-pays buckets
- Allow that service-account to access the _release_ bucket on the parent project.

### DIFF
--- a/stack/__main__.py
+++ b/stack/__main__.py
@@ -18,6 +18,9 @@ import pulumi_gcp as gcp
 
 DOMAIN = 'populationgenomics.org.au'
 CUSTOMER_ID = 'C010ys3gt'
+BILLING_ACCOUNT_ID = '01D012-20A6A2-CBD343'
+BILLING_PROJECT_ID = 'billing-admin-290403'
+
 REGION = 'australia-southeast1'
 ANALYSIS_RUNNER_PROJECT = 'analysis-runner'
 CPG_COMMON_PROJECT = 'cpg-common'
@@ -60,6 +63,7 @@ def main():  # pylint: disable=too-many-locals,too-many-branches
     # Fetch configuration.
     config = pulumi.Config()
     enable_release = config.get_bool('enable_release')
+    enable_egress_project = config.get_bool('enable_egress_project')
     archive_age = config.get_int('archive_age') or 30
 
     dataset = pulumi.get_stack()
@@ -629,6 +633,7 @@ def main():  # pylint: disable=too-many-locals,too-many-branches
         member=pulumi.Output.concat('group:', access_group.group_key.id),
     )
 
+    release_bucket = None
     if enable_release:
         release_bucket = create_bucket(
             bucket_name('release-requester-pays'),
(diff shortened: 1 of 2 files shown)


stack/__main__.py
```diff
@@ -650,6 +657,117 @@ def add_access_group_cache_as_secret_member(secret, resource_prefix: str):
...
 
+    if enable_egress_project:
+        if not release_bucket:
+            raise ValueError(
+                'Requested egress project, but no bucket is available to egress from'
+            )
+
+        egress_buckets = {'release': release_bucket}
+
+        project_name = f'cpg-{dataset}-egress'
```
lgruen, line 668: How about we make this a bit more generic, so we can potentially use this for compute as well in the future? Can we make the name consistent with the existing project names?

```suggestion
        project_name = f'{project_id}-shared'
```

stack/__main__.py
```diff
@@ -650,6 +657,117 @@ def add_access_group_cache_as_secret_member(secret, resource_prefix: str):
             member=pulumi.Output.concat('group:', release_access_group.group_key.id),
         )
 
+    if enable_egress_project:
+        if not release_bucket:
+            raise ValueError(
+                'Requested egress project, but no bucket is available to egress from'
+            )
```
lgruen, line 664: We could skip that, as the same logic will apply once we share `main` buckets (with Requester Pays enabled).

lgruen (changes requested):

stack/__main__.py
illusional: Are you okay if I leave this here until that's implemented, already caught me once :)

stack/__main__.py
lgruen: Do you want to rename all the other "egress" occurrences as well?

lgruen (changes requested):
Do you want to rename all the other "egress" occurrences as well?

stack/__main__.py
illusional: Ah right, I see sorry.

illusional: Tested and pushed new changes :)

stack/__main__.py
```diff
@@ -650,6 +657,117 @@ def add_access_group_cache_as_secret_member(secret, resource_prefix: str):
             member=pulumi.Output.concat('group:', release_access_group.group_key.id),
         )
 
+    if enable_shared_project:
+        if not release_bucket:
+            raise ValueError(
+                'Requested shared project, but no bucket is available to shared from'
```
lgruen, line 663:
```suggestion
                'Requested shared project, but no bucket is available to share'
```

stack/__main__.py
```diff
@@ -650,6 +657,117 @@ def add_access_group_cache_as_secret_member(secret, resource_prefix: str):
...
+        shared_service_account = gcp.serviceaccount.Account(
+            'budget-shared-service-account',
+            account_id=f'shared',
+            opts=pulumi.resource.ResourceOptions(
+                depends_on=[shared_cloudidentity],
+            ),
+            project=shared_project.name,
+        )
+
+        # Allow the usage of requester-pays buckets.
```
lgruen, line 751:
```suggestion
        # Allow the usage of Requester Pays buckets.
```

lgruen (approved):

stack/__main__.py
```diff
@@ -650,6 +657,117 @@ def add_access_group_cache_as_secret_member(secret, resource_prefix: str):
...
+        )
+        cloudbillingbudgets = gcp.projects.Service(
+            'cloudbillingbudgets-service',
+            service='billingbudgets.googleapis.com',
+            disable_on_destroy=False,
+            project=project_id,
+            opts=pulumi.resource.ResourceOptions(depends_on=[cloudbilling]),
+        )
+
+        # create project to cover shared
```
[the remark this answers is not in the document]
illusional: Sorry, this was sloppy from me

lgruen:
As this is the first time we're creating `release` buckets, should we use this opportunity to remove the `-requester-pays` suffix from the bucket name if we'll be enabling Requester Pays on all buckets anyhow?

illusional:
Addressed requester-pays release bucket naming in another PR, merging this for now :)

PR #476, opened 2022-08-24
Add shared project + budget for seqr release

illusional:
Requires https://github.com/populationgenomics/analysis-runner/pull/473


cassimons:
To make sure I am following, is this a stop-gap until we can convert the main bucket to requester pays? Is the process is:
- It creates a new requester pays bucket
- Files are duplicated to this bucket as needed
- A service account is created with read access and a small budget
- Service account is provided to the collaborator to do the downloads

Are we using any sort of TTL on the copies or will we need to do manual cleanups?

illusional:
Hey @cassimons, correct, copying to a release bucket is a stop gap with no explicit TTL. The intention is to leave them in main bucket but currently this is not possible.

We control egress cost through the new "shared project", and we will generate SA credentials for the collaborator to download the data.

We can chat more on slack, but maybe we can get them to notify us when they've finished downloading it.

lgruen (approved):

PR #480, opened 2022-08-25
Update path match to be case sensitive

MattWellie:
e.g. https://batch.hail.populationgenomics.org.au/batches/163811/jobs/1

Analysis-runner was able to start this job, despite the provided script name `plot_alt_af.py` not being a case-specific match to the real name `plot_alt_AF.py`
The cause is (likely) the lack of case recognition in ZSH, which results in case insensitive matching being done by `os.path.exists`

Playing with this locally in ZSH:
```bash
ls scripts/rv_expression_association/plot/plot_alt_AF.py
> scripts/rv_expression_association/plot/plot_alt_AF.py
ls scripts/rv_expression_association/plot/plot_alt_af.py
> scripts/rv_expression_association/plot/plot_alt_af.py
```


lgruen (changes requested):
You'll probably also want to add a version bump.

vladsavelyev (approved):

lgruen (approved):

PR #481, opened 2022-08-25
Fix analysis request convert `frozendict` from read_configs to `dict`

vladsavelyev:
`frozendict` is not serialisable and cannot be passed via a POST request, so the analysis runner with `--config` option fails.


lgruen (approved):
Does this need a version bump?

PR #482, opened 2022-08-25
Add rare-disease to seqr allowed-repos

illusional:
To move seqr files.

Edit to run: https://github.com/populationgenomics/rare-disease/pull/10


lgruen (approved):
Move which files where? ;)

PR #490, opened 2022-08-31
Revert "Fix fetching of ID tokens when using a service account (#487)"

lgruen:


MattWellie (approved):
I thought it would be more intuitive to just do a full revert (i.e. back to version 2.32.4, instead of 2.32.4 being the same as 2.32.6)

illusional (approved):

lgruen:
> I thought it would be more intuitive to just do a full revert (i.e. back to version 2.32.4, instead of 2.32.4 being the same as 2.32.6)

But 2.32.5 has already been released, so we'd need to yank that version?

lgruen:
> But 2.32.5 has already been released, so we'd need to yank that version?

I'm not sure how our automatic analysis-runner "please update" messages would handle that.

illusional:
> I'm not sure how our automatic analysis-runner "please update" messages would handle that.

Not very well: https://github.com/populationgenomics/analysis-runner/blob/43cb4ce72c4a8601846a453a4b8c92821926fe32/analysis_runner/util.py#L147

PR #493, opened 2022-09-02
Fix billing custom period for shared projects

illusional:
Missed in the last PR because I had unpushed commits


lgruen (approved):
Did you already check that this doesn't break the billing report and still triggers unlinking the billing account when a project goes over budget?

illusional:
> Did you already check that this doesn't break the billing report and still triggers unlinking the billing account when a project goes over budget?

This is how I noticed the changes weren't included in my previous PR. I've confirmed that the requester pays works and the charges do fall onto the correct project (through bigquery). I've dropped the budget down, ran some egress and waiting for it to appear. It hasn't happened immediately - so I hope it doesn't take a long time to appear.

PR #499, opened 2022-09-06
Add list of repos that are allowed for all datasets

lgruen:
As suggested by @MattWellie

### DIFF
--- a/tokens/main.py
+++ b/tokens/main.py
@@ -70,7 +70,7 @@ def main():
     for dataset, allowed_repos in ALLOWED_REPOS.items():
         entries = {
             'projectId': get_project_id(dataset),
-            'allowedRepos': ALWAYS_ALLOWED_REPOS + allowed_repos,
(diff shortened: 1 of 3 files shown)


tokens/main.py
```diff
@@ -63,7 +68,10 @@ def main():
 
     config = {}
     for dataset, allowed_repos in ALLOWED_REPOS.items():
-        entries = {'projectId': get_project_id(dataset), 'allowedRepos': allowed_repos}
+        entries = {
+            'projectId': get_project_id(dataset),
+            'allowedRepos': ALWAYS_ALLOWED_REPOS + allowed_repos,
```
illusional, line 73: To avoid dupes, not a big deal either way though.
```suggestion
            'allowedRepos': list(set(ALWAYS_ALLOWED_REPOS + allowed_repos)),
```

illusional (approved):

tokens/main.py
lgruen: Good catch, thanks!

PR #500, opened 2022-09-06
Bump version: 2.32.11 → 2.33.0: Allow launching of non-preemptible jobs

lgruen:


analysis_runner/cli_analysisrunner.py
```diff
@@ -57,6 +57,14 @@ def add_analysis_runner_args(parser=None) -> argparse.ArgumentParser:
         ),
     )
 
+    parser.add_argument(
+        '--preemptible',
+        required=False,
+        action=argparse.BooleanOptionalAction,
+        default=True,
```
illusional, line 64: How does default true here work if it's not a flag, `--preemptible false`?

server/main.py
```diff
@@ -153,6 +154,9 @@ async def index(request):
         job.cpu(cpu)
     if memory:
         job.memory(memory)
+    if preemptible is not None:
```
illusional, line 157: Should we default to True here if `None` to be consistent with the CLI description?

illusional (approved):
Minor comments but looks good!

analysis_runner/cli_analysisrunner.py
lgruen: As far as I understand, it should understand both `--preemptible` and `--no-preemptible` and set the value accordingly.

server/main.py
lgruen: Good idea! I've added a default value when fetching the value from the dictionary.

PR #503, opened 2022-09-07
Compute disk usage stats recursively for all folders

lgruen:
(Except for `.ht` / `.mt`.)

I'd like to build a simple treemap visualization based on this.

### DIFF
--- a/disk_usage/disk_usage.py
+++ b/disk_usage/disk_usage.py
@@ -82,8 +82,8 @@ def main():
 
     output = output_path('disk_usage.json.gz')
     logging.info(f'Writing results to {output}...')
-    with AnyPath(output).open('w') as f:
-        with gzip.GzipFile(fileobj=f, mode='w') as gzf:
+    with AnyPath(output).open('wb') as f:
+        with gzip.open(f, 'wt') as gzf:
             json.dump(aggregate_stats, gzf)
 
 
--- a/disk_usage/disk_usage.py
+++ b/disk_usage/disk_usage.py
@@ -3,6 +3,7 @@
 """Computes aggregate bucket disk usage stats."""
 
 from collections import defaultdict
+import gzip
 import json
 import logging
 from cloudpathlib import AnyPath
@@ -79,10 +80,11 @@ def main():
 
         logging.info(f'{bucket_name} contains {count} blobs.')
 
-    output = output_path('disk_usage.json')
+    output = output_path('disk_usage.json.gz')
     logging.info(f'Writing results to {output}...')
     with AnyPath(output).open('wt') as f:
-        json.dump(aggregate_stats, f)
+        with gzip.open(f, 'wt') as gzf:
+            json.dump(aggregate_stats, gzf)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


disk_usage/disk_usage.py
```diff
@@ -27,40 +28,33 @@
...
 
-    aggregate_stats = defaultdict(AggregateStats)
+    aggregate_stats = defaultdict(lambda: defaultdict(int))
     for bucket_suffix in BUCKET_SUFFIXES:
+        if access_level == 'test' and not bucket_suffix.startswith('test'):
+            continue  # Skip main buckets when testing.
+
         bucket_name = f'cpg-{dataset}-{bucket_suffix}'
         logging.info(f'Listing blobs in {bucket_name}...')
         blobs = storage_client.list_blobs(bucket_name)
```
illusional, line 60: Does this not already list every single file in the bucket? I thought you had to be explicit about filtering to just the bucket root or something.

disk_usage/disk_usage.py
```diff
@@ -27,40 +28,33 @@
 
 
 def aggregate_level(name: str) -> str:
-    """Returns a prefix for the given blob name at the aggregation level."""
+    """Returns a prefix for the given blob name at the folder or Hail table level."""
```
illusional, line 31: Should throw a couple docstring tests in here :)

disk_usage/disk_usage.py
```diff
@@ -69,23 +63,28 @@ def main():
...
-            stats.num_blobs += 1
+            folder = aggregate_level(blob.name)
+            last_index = 0
+            while True:
+                index = folder.find('/', last_index)
+                substr = folder[:index] if index != -1 else folder
+                path = f'gs://{bucket_name}/{substr}'
+                stats = aggregate_stats[path]
+                stats['size'] += blob.size
+                stats['num_blobs'] += 1
```
illusional, line 74: Stylistically I prefer setting on the instance, but interested for this many millions if it makes a substantial difference.

```suggestion
                aggregate_stats[path]['size'] += blob.size
                aggregate_stats[path]['num_blobs'] += 1
```

disk_usage/disk_usage.py
```diff
@@ -27,40 +28,33 @@
...
 
 def aggregate_level(name: str) -> str:
-    """Returns a prefix for the given blob name at the aggregation level."""
+    """Returns a prefix for the given blob name at the folder or Hail table level."""
     ht_index = name.find('.ht/')
     if ht_index != -1:
         return name[: ht_index + 3]
     mt_index = name.find('.mt/')
     if mt_index != -1:
         return name[: mt_index + 3]
```
illusional, line 37: You could probably use the [walrus syntax](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions) here:

```suggestion
    if (ht_index := name.find('.ht/')) != -1:
        return name[: ht_index + 3]
    if (mt_index := name.find('.mt/')) != -1:
        return name[: mt_index + 3]
```

illusional (approved):

disk_usage/disk_usage.py
lgruen: Probably doesn't make a huge difference overall.

disk_usage/disk_usage.py
lgruen: Thanks, done!

disk_usage/disk_usage.py
lgruen: Nice, thanks!

PR #504, opened 2022-09-07
Allow new_stack.py to create batch billing projects

illusional:
Considered the "ask for forgiveness" approach but prefer requesting the project up front, even if it involves one extra request.

### DIFF
--- a/stack/new_stack.py
+++ b/stack/new_stack.py
@@ -58,10 +58,22 @@ ORGANIZATION_ID = '648561325637'
 BILLING_ACCOUNT_ID = '01D012-20A6A2-CBD343'
 BILLING_PROJECT_ID = 'billing-admin-290403'
 HAIL_PROJECT = 'hail-295901'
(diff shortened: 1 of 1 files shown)


stack/new_stack.py
```diff
@@ -373,17 +444,24 @@ def _create_hail_service_account(username, hail_auth_token):
     post_resp.raise_for_status()
 
 
+def _get_standard_hail_account_names(dataset):
+    username_suffixes = ['-test', '-standard', '-full']
+    return [dataset + suffix for suffix in username_suffixes]
```
lgruen, line 449:
```suggestion
    return ['aggregate-billing'] + [dataset + suffix for suffix in username_suffixes]
```

lgruen (approved):
It's so nice to see this fully automated now. Great work!

stack/new_stack.py
illusional: This is used to generate the hail accounts, the billing aggregator username is added when setting up the hail batch billing project.

PR #506, opened 2022-09-08
Fix aggregation at the bucket root level

lgruen:


disk_usage/disk_usage.py
```diff
@@ -70,19 +70,14 @@ def main():
...
             if count % 10**6 == 0:
                 logging.info(f'{count // 10**6} M blobs...')
-            folder = aggregate_level(blob.name)
-            last_index = 0
+            folder = f'/{aggregate_level(blob.name)}'
             while True:
-                index = folder.find('/', last_index)
-                substr = folder[:index] if index != -1 else folder
-                path = f'gs://{bucket_name}/{substr}'
+                path = f'gs://{bucket_name}{folder}'
```
illusional, line 75: Not `os.path.join`? Then you wouldn't need the leading slash

illusional (approved):

disk_usage/disk_usage.py
lgruen: I added the leading slash to simplify the `.find('/')` logic a bit -- otherwise there's an edge case at the root level that would require extra logic (which I got wrong before 😉).

PR #509, opened 2022-09-09
Remove `nagim` and `amp-pd` stacks

lgruen:


tokens/repository-map.json
```diff
@@ -1,146 +1,83 @@
...
-    ],
-    "validation": [
-        "automated-interpretation-pipeline",
-        "production-pipelines",
-        "joint-calling",
-        "gatk-sv",
-        "rare-disease",
-        "large-cohort-pipeline"
-    ]
+  "acute-care": [
```
illusional: Can we standardise our json formatting? Looks like everyone does it differently and this file gets changed frequently.

illusional (approved):

tokens/repository-map.json
lgruen: Sorry, that was before I reverted the VS Code's auto-format (should be better now).

Probably `new_stack.py` applies different formatting than VS Code?

PR #512, opened 2022-09-13
Finish cromwell watcher loop on success

vladsavelyev:
Cromwell watcher would still loop indefinitely when workflow finished successfully. Adding a `break` to fix it


vivbak (approved):
🙏 amazing tysm 🙏

PR #513, opened 2022-09-13
improves flexibility of MT copying script

MattWellie:
Initial version didn't allow for 'no samples', i.e. retain all original samples in the copied data
Initial version didn't accept the hail syntax for `chr1:start-end` (or mixing `start-12345`, or `7890-end`)
alters the input validation to take these into account

### DIFF
--- a/scripts/subset_matrix_table.py
+++ b/scripts/subset_matrix_table.py
@@ -220,7 +220,7 @@ if __name__ == '__main__':
     main(
         mt_path=args.i,
         output_root=args.out,
-        samples=set(args.s),
+        samples=set(args.s) if args.s else None,
         out_format=args.format,
         locus=locus_interval,
         keep_hom_ref=args.keep_ref,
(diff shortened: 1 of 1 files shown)


annacuomo (approved):

scripts/subset_matrix_table.py
```diff
@@ -140,28 +140,41 @@ def clean_locus(contig: str, pos: str) -> hl.IntervalExpression | None:
...
-            end += 1
     else:
+        assert int(
+            pos
+        ), f'if only one position is specified, it must be numerical: {pos}'
         start = int(pos)
         end = start + 1
 
-    return hl.parse_locus_interval(f'{contig}:{start}-{end}')
+    return hl.parse_locus_interval(f'{contig}:{start}-{end}', reference_genome='GRCh38')
```
illusional, line 177: You've dropped the `start < end` check if they're both ints which would be useful IMO to keep.

illusional (approved):
One minor nitpick :)

scripts/subset_matrix_table.py
MattWellie: Updated to re-add this, thanks!

PR #516, opened 2022-09-23
Table version of the MT subset script

MattWellie:
Adds a second script, which copies a Hail Table. Mostly copied from the MatrixTable subset script, but removes all references to samples/genotypes

### DIFF
--- a/scripts/subset_hail_table.py
+++ b/scripts/subset_hail_table.py
@@ -1,0 +1,169 @@
+#!/usr/bin/env python3
+
+"""
+Takes a path to a Hail Table and an output name
+Writes data into the test bucket for the dataset
+
+Optionally a locus can be provided to reduce the output
+Optionally supply both --chr and --pos; subset to a specific locus
+The pos format can be a single int, or a "start-end"
+"""
+
+from argparse import ArgumentParser
+import logging
+import sys
+
+import hail as hl
+
+from cpg_utils.hail_batch import output_path, init_batch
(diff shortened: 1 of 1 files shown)


annacuomo (approved):

scripts/subset_hail_table.py
```diff
@@ -0,0 +1,169 @@
...
+if __name__ == '__main__':
+
+    logging.basicConfig(
+        level=logging.INFO,
+        format='%(asctime)s %(levelname)s %(module)s:%(lineno)d - %(message)s',
+        datefmt='%Y-%M-%d %H:%M:%S',
+        stream=sys.stderr,
+    )
+
+    parser = ArgumentParser()
```
illusional, line 140: We generally use click for CLI, I think it's a bit cleaner. Not a big deal for this PR, but for new scripts :)

illusional (commented):
If you provide no params, this won't filter the table though right? Could we ensure at least one filter is provided to avoid a straight copy?

MattWellie:
@illusional There are times when a full copy would be useful, especially for smaller tables. Definitely something we should be mindful of in terms of storage costs if people are able to generate massive copies

MattWellie:
Thanks @illusional, linting issues resolved. 

I'm going to hit merge here as @lgruen is at the team-leads meeting, but I'll calendar something to discuss this another time (as this + the MT copying script give users ability to copy mad amounts of data around, which should maybe be limited, or at least discussed)

PR #525, opened 2022-10-04
Remove stack + related files

illusional:
All migrated to https://github.com/populationgenomics/cpg-infrastructure
I've caught all latest changes, and ready for a new PR over in the new home.


lgruen (approved):
Woohoo! Might need a follow-up update of the README as well?

PR #532, opened 2022-10-12
Build `latest` tag for driver image

vladsavelyev:
Context: https://github.com/populationgenomics/cpg-utils/pull/100#discussion_r993025629


.github/workflows/deploy_server.yaml
```diff
@@ -49,23 +50,25 @@ jobs:
...
 
     - name: "build server image"
       run: |
-        docker build --build-arg DRIVER_IMAGE=$DRIVER_IMAGE --tag $SERVER_IMAGE server
+        docker build --build-arg DRIVER_IMAGE=$DRIVER_IMAGE:$IMAGE_TAG --tag $SERVER_IMAGE:$IMAGE_TAG server
 
     - name: "push server image"
       run: |
-        docker push $SERVER_IMAGE
+        docker push $SERVER_IMAGE:$IMAGE_TAG
```
lgruen, line 67: Should we also have `latest` here for consistency?

lgruen (approved):
Thanks!

.github/workflows/deploy_server.yaml
vladsavelyev: Done!

PR #533, opened 2022-10-21
script for copying clinvar to CPG reference

MattWellie:
Clinvar annotations created by the Broad are written to gs://seqr-reference-data/GRCh38/clinvar/clinvar.GRCh38.{date}.ht

This script is for copying their data to our cpg-reference bucket path defined here:
https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/config-template.toml#L228-L231
(the name of the latest clinvar file will need to be updated in this config template)

The date is an argument value so that we don't need a new script for each new version
Output path is also configurable, but with the default that our current `cpg_utils.hail_batch.reference_path('seqr/clinvar'))` will generate


scripts/copy_clinvar.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# this should be run with high-core low memory resources
+# e.g.
+# analysis-runner analysis-runner --cpu 8 --memory 2Gi scripts/copy_clinvar.sh 2022-09-17
```
vladsavelyev, line 5: Should we try `--cpu 16 --memory lowmem`?

scripts/copy_clinvar.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -ex
+
+DATE=${1}
+CPG_ANNO=${2:-"gs://cpg-reference/seqr/v0-1/"}
```
vladsavelyev, line 10: Let's maybe drop this `v0-1` version tag, since we are using the Broad's datestamp versions?

vladsavelyev (approved):
Thank you Matt for setting this up!

Two suggestions :)

Also, the dataset we should generally be using to run such script is `reference`, but I'm not sure if we finished setting it up

scripts/copy_clinvar.sh
MattWellie: updated the docstring

https://hail.is/docs/batch/api/batch/hailtop.batch.job.Job.html#hailtop.batch.job.Job.memory

MattWellie:
@vladsavelyev  I've updated the docstring to resemble what should be a complete command

PR #534, opened 2022-10-21
update clinvar copy

MattWellie:
`gcloud storage` wasn't any good - error message advice was to reference the alpha featureset

also checks that a non-null argument was provided

### DIFF
--- a/scripts/copy_clinvar.sh
+++ b/scripts/copy_clinvar.sh
@@ -14,7 +14,13 @@
 
 set -ex
 
+if [ -z "$1" ]
+  then
+    echo "No date argument supplied"
+    exit1
+fi
+
 DATE=${1}
 CPG_ANNO=${2:-"gs://cpg-reference/seqr/"}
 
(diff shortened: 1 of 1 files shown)


scripts/copy_clinvar.sh
```diff
@@ -14,7 +14,13 @@
...
+  then
+    echo "No date argument supplied"
+    exit1
+fi
+
 DATE=${1}
 CPG_ANNO=${2:-"gs://cpg-reference/seqr/"}
 
-gcloud storage cp -r "gs://seqr-reference-data/GRCh38/clinvar/clinvar.GRCh38.${DATE}.ht" "${CPG_ANNO}"
+gcloud alpha storage cp -r "gs://seqr-reference-data/GRCh38/clinvar/clinvar.GRCh38.${DATE}.ht" "${CPG_ANNO}"
```
lgruen, line 26: I wonder if a `gcloud -q components update` would fetch a newer version, so we don't have to use `alpha`? Should hopefully be easy to verify by running the driver image in Docker, e.g. in a Cloud Shell.

lgruen (approved):

PR #538, opened 2022-10-24
Fix XML library error and add commonly used R packages

lgruen:


KatalinaBobowik (approved):

driver/Dockerfile.r
```diff
@@ -16,9 +16,9 @@ RUN apt update && \
         r-argparser \
         r-arrow \
         r-base \
+        r-biomartr \
```
KatalinaBobowik, line 19: do we need biomartr?

lgruen: We certainly don't need it. Just depends on whether we think it's used often enough to justify a slightly larger image size.

KatalinaBobowik: I'm just saying that `biomartr` is a different package than `biomaRt`, which I'm using in my current scrip. I haven't used `biomartr` but seems to be quite similar, albeit a different syntax

lgruen: Thanks! Fixed in https://github.com/populationgenomics/analysis-runner/pull/539.

PR #541, opened 2022-10-27
Add script to copy crams/gvcfs for a family to test

cassimons:
Adds a simple that given a project ID and one or more family IDs, copies cram and gvcf files for each individual in the family into the project test bucket.

### DIFF
--- a/scripts/copy_family_to_test.py
+++ b/scripts/copy_family_to_test.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+
+"""
+
+Given a project ID and one or more family IDs, copies cram and gvcf
+files for each individual in the family into the project test bucket.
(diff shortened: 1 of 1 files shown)


MattWellie (approved):

vladsavelyev:
Just noting that our [test subset creation script](https://github.com/populationgenomics/sample-metadata/blob/dev/scripts/create_test_subset.py) copies GVCF and CRAM into test as well, but it would use different CPG IDs for the sample samples (because it also creates `Sample` and `Analysis` entries), so we could end up with some extra clutter in the `-test` buckets. Perhaps not a big deal, both scripts are idempotent so the bucket can be cleaned at any time and scripts can be rerun at any time if needed as well.

PR #543, opened 2022-11-03
Relax pinned versions to be consistent with the hail package

lgruen:
Related to https://github.com/populationgenomics/production-pipelines/pull/174, assuming it's based on the issue described in https://github.com/pypa/pip/issues/8713.


vladsavelyev:
Can we remove pins for all deps (or keep the minimum version pins)? It might quickly go outdated and cause problems for packages that depend on `analysis-runner`.

vladsavelyev (approved):

setup.py
```diff
@@ -36,18 +36,18 @@ def read_file(filename: str) -> str:
...
+        'google-api-python-client>=2.10.0,<3',
+        'google-auth>=1.27.0,<2',
+        'google-cloud-logging>=3.1.1,<4',
+        'google-cloud-pubsub>=2.3.0,<3',
+        'google-cloud-secret-manager>=2.2.0,<3',
+        'google-cloud-storage==1.25.*',
         'Jinja2==3.0.3',
         'kubernetes',
-        'protobuf==3.20.1',
+        'protobuf==3.20.2',
```
vladsavelyev, line 47: Did you mean to keep the hard pin for `protobuf`?

setup.py
```diff
@@ -36,18 +36,18 @@ def read_file(filename: str) -> str:
...
-        'google-cloud-logging==3.1.1',
-        'google-cloud-pubsub==2.3.0',
-        'google-cloud-secret-manager==2.2.0',
-        'google-cloud-storage==1.25.0',
+        'google-api-python-client>=2.10.0,<3',
+        'google-auth>=1.27.0,<2',
+        'google-cloud-logging>=3.1.1,<4',
+        'google-cloud-pubsub>=2.3.0,<3',
+        'google-cloud-secret-manager>=2.2.0,<3',
+        'google-cloud-storage==1.25.*',
```
vladsavelyev, line 44: Did you mean to pin the minor version here?

lgruen: That's what I'd usually do, but this is copied directly from `hail` for maximum compatibility.

setup.py
lgruen: Same here, this comes directly from `hail`, to avoid conflicts.

lgruen:
> Can we remove pins for all deps (or keep the minimum version pins)? It might quickly go outdated and cause problems for packages that depend on analysis-runner.

Restricting major versions does provide some protection against breaking changes, but maybe I'm trying to be as compatible with `hail`'s pinned packages as possible here.

setup.py
vladsavelyev: Shouldn't `pip` automatically pull the requirements that `hail` needs from `hail`'s setup.py?

lgruen: Yes, you're right! Changed.

setup.py
lgruen: Updated.

lgruen:
Now updated to only enforce minimum versions, as the rest should come automatically from `hail`.

PR #544, opened 2022-11-03
Simplify deps

lgruen:
`setup.py` should only list what's actually needed for the package.


vladsavelyev (approved):
I like it! I will pull everything needed for `hail` recursively based on `hail` requirements.

PR #545, opened 2022-11-03
Driver: move installing `hail` from source before `analysis-runner`

vladsavelyev:
Thanks @illusional for the idea. It does seem to help with `pip` pulling older `hail` versions. 

However, building hail from source doesn't work for yet another reason:

```
...
#5 211.0 Starting a Gradle Daemon (subsequent builds will be faster)
#5 305.9 > Task :compileJava NO-SOURCE
#5 335.9 > Task :compileScala
#5 652.5 > Task :compileScala FAILED
#5 653.8
#5 653.8 FAILURE: Build failed with an exception.
#5 653.8
#5 653.8 * What went wrong:
#5 653.8 Execution failed for task ':compileScala'.
#5 653.8 > Failed to run Gradle Worker Daemon
#5 653.8    > Process 'Gradle Worker Daemon 1' finished with non-zero exit value 137
...
```

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -16,7 +16,8 @@
 
 disable=f-string-without-interpolation,logging-fstring-interpolation,inherit-non-class,
             too-few-public-methods,fixme,line-too-long,too-many-statements,
-            too-many-lines,too-many-instance-attributes,R0801,c-extension-no-member,no-member,W3101
+            too-many-lines,too-many-instance-attributes,R0801,c-extension-no-member,
+            no-member,W3101,import-error,no-name-in-module
 
 
 # Overriding variable name patterns to allow short 1- or 2-letter variables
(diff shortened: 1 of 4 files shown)


lgruen:
IIRC, the idea to run `pip` after compiling Hail was to avoid `pip` pulling any `hail` package in the first place.

Instead of running `make install`, do we need to build the wheel and install that, so `pip` recognizes the installation correctly?

lgruen:
Good idea, as hopefully that should avoid `pip` trying to download any `hail` package, assuming that the local installation is recognized by `pip`.

I'm trying to reproduce that compilation failure locally.

lgruen:
Your change "just works" for me locally. I think `exit value 137` might indicate an OOM issue? Maybe we can just increase the requested memory for that GitHub Action to avoid this or alternatively request less parallelism for the build?

vladsavelyev:
@lgruen I actually tried building locally, so OOM makes sense :D let's merge and ci if CI works well

lgruen (approved):

lgruen:
Great! Now running here: https://github.com/populationgenomics/analysis-runner/actions/runs/3389795997

PR #546, opened 2022-11-03
Add indexes to copy to test  script

cassimons:


MattWellie (approved):
Not specifically related to this change, but it might also be worth checking that the files exist before initiating the copy

PR #549, opened 2022-11-07
Fix dataproc init script

vladsavelyev:
Harmonises the init script with the driver dockerfile and fixes the numpy/scipy installation error


lgruen (approved):
🙇

PR #551, opened 2022-11-07
First commit for adding VEP batch backend script

KatalinaBobowik:
This is still in an unfinished state, as we still need to add a config file, but would love any feedback regarding the VEP script and README. @MattWellie if you can comment/make suggestions on lines 33 and 35 of the `vep_batch_backend.py` script, that would be much appreciated (or really anything else you spot).


vladsavelyev (approved):

scripts/vep_backend/vep_batch_backend.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Run VEP in parallel using Picard tools intervals as partitions.
+    """
+
+    b = get_batch(f'Run VEP with Batch Backend, VEP v{vep_version}')
+    add_vep_jobs(
+        b=b,
+        # I'm not sure what 'to_path' does here - is it needed?
+        vcf_path=to_path(dataset_path(vcf_path)),
+        # would something like this work to get the tmp path from the output_path input?
```
MattWellie, line 35: in your example run command from the README I think this would deliver 

`gs://cpg-tx-adapt-test-tmp/vep_batch/v0/vcf_fragments`

scripts/vep_backend/vep_batch_backend.py
```diff
@@ -0,0 +1,44 @@
...
+)
+def main(vep_version: str, vcf_path: str, output_ht: str, scatter_count: int):
+    """
+    Run VEP in parallel using Picard tools intervals as partitions.
+    """
+
+    b = get_batch(f'Run VEP with Batch Backend, VEP v{vep_version}')
+    add_vep_jobs(
+        b=b,
+        # I'm not sure what 'to_path' does here - is it needed?
```
MattWellie, line 33: `to_path` converts a string path location to either a Path.PosixPath (if it's local), or a cloudpathlib.CloudPath (if it's cloud). These Path objects have built-in methods like `Path.suffix()` and `Path.exists()`, so generally a bit more useful than the plain strings. `add_vep_jobs` expects these Path objects instead of plain strings

scripts/vep_backend/vep_batch_backend.py
KatalinaBobowik: Yeah, I just don't want to add too many input parameters, so thought it'd be nice to 'elegantly' use something from the input we already have

MattWellie:
AFAIK these values are the only ones included in the runtime config by default in analysis-runner
https://github.com/populationgenomics/analysis-runner/blob/main/server/main.py#L105-L123

I've added a config file with a few more entries, locating VEP resources and identifying the annotation mount location. I also moved the scatter_count there, and the image name is published in the `get_batch()` name instead of taking an argument value.

scripts/vep_backend/vep_batch_backend.py
```diff
@@ -0,0 +1,37 @@
...
+)
+def main(vcf_path: str, output_ht: str):
+    """
+    Run VEP in parallel using Picard tools intervals as partitions.
+    """
+    vep_image = get_config()['images']['vep']
+    b = get_batch(f'Run VEP with Batch Backend, image {vep_image}')
+    add_vep_jobs(
+        b=b,
+        vcf_path=to_path(dataset_path(vcf_path)),
```
MattWellie, line 28: Thinking about this, it could be better to pass this as an explicit path?

KatalinaBobowik: Yes, good point. In case the datasets are different. I'll change that

MattWellie (approved):
Probably worth running this through in `test` before merging, just in case the config is missing any information. Thanks for putting this together!

scripts/vep_backend/vep_batch_backend.py
vladsavelyev: Thanks Matt for a really good explanation!

I'd also add that for a function to expect a `Path` object instead of random `str` helps to make the code more robust: static code analysis tools (like `mypy` we use in the pre-commit linting toolset) can catch early if by a mistake a random string is passed into the function, or a path that is not correctly formed.

PR #554, opened 2022-11-09
Fix invalid config (server side)

illusional:
CLI supplies this as null which breaks because it's present just not value
No bumpversion as only server change

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -98,8 +98,8 @@ async def index(request):
 
     # Prepare the job's configuration and write it to a blob.
     config = get_baseline_config(server_config, dataset, access_level, output_prefix)
-    if 'config' in params:  # Update with user-specified configs.
-        update_dict(config, params['config'])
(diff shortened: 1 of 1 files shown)


vladsavelyev (approved):

server/main.py
```diff
@@ -98,8 +98,8 @@ async def index(request):
 
     # Prepare the job's configuration and write it to a blob.
     config = get_baseline_config(server_config, dataset, access_level, output_prefix)
-    if 'config' in params:  # Update with user-specified configs.
-        update_dict(config, params['config'])
+    if user_config := params.get('config'):  # Update with user-specified configs.
```
lgruen, line 101: Does this need to be fixed on the Cromwell side as well?

illusional: Ah yes I'd assume so, to be honest I didn't even realise Cromwell used the config, but makes sense. Will send a PR

PR #559, opened 2022-11-25
Merge storage/images/references templates for the baseline config

vladsavelyev:
* Reads `images`, `references`, `storage` from `gs://cpg-config/templates` and adds them into the baseline config, below the user config.
* It's not cloud-agnostic as we hardcode the infra name `gcp`, however as far as I understand AR is pretty much GCP-specific at the moment.
* Few minor changes: 
  * rename `config` to `run_config` to distinguish from `server_config`,
  * remove dependency on `cpg_utils.hail_batch.Namespace` which is dropped in https://github.com/populationgenomics/cpg-utils/pull/117


analysis_runner/dataproc.py
```diff
@@ -210,6 +208,7 @@ def _add_start_job(  # pylint: disable=too-many-arguments
 
     # Note that the options and their values must be separated by an equal sign.
     # Using a space will break some options like --label
+    namespace = 'test' if ACCESS_LEVEL == 'test' else 'main'
```
lgruen, line 211: It's a simple check, but we duplicate this in a number of places (e.g. also `server/util.py` below). Is it worth putting this into a tiny shared function (either in the analysis-runner or `cpg-utils`)?

lgruen (approved):
Awesome, thanks!

@illusional FYI

analysis_runner/dataproc.py
illusional: +1 to this!!

vladsavelyev: It actually removed something like that in the last cpg-utils PR 😅 But having a `Namespace` enum was an overkill anyway, agree that a simple function would be better

PR #563, opened 2022-11-30
allow driver override in config

### LINKED ISSUE #562: DRIVER_IMAGE in conf
see: https://github.com/populationgenomics/analysis-runner/blob/5f6536ce7828bf06973333bcd7a1e366962f06ee/server/util.py#L187

When the analysis-runner driver image is overridden using `--image`, the 'driver_image' in the toml config is still the canonical driver image, with no option to override. The metadata for the job _does_ assign driverImage correctly.

Would it make sense that the image 'currently driving' the batch is persisted into config as the `driver`? I don't mind making the PR if we want to change the behaviour, but I wanted to check if this is intentional.

A use case I'm considering is that from the driver jobs, I expect `get_config()['workflow']['driver_image']` to provide the docker image which is currently being run. For non-default driver images this doesn't hold true.

MattWellie:
Closes #562 

Driver is optional (changing driver is not permitted via cromwell, which shares this method)


illusional (changes requested):
Can you action the linter, there's a similar flow in Cromwell that just needs to be passed `driver = None`, and there are merge conflicts

PR #565, opened 2022-11-30
Allow driver conf override

### LINKED ISSUE #562: DRIVER_IMAGE in conf

MattWellie:
prev. PR developed an incorrect diff vs. main

Closes #562

If a driver image is set using `--image` this will propagate to the config TOML as the driver_image


illusional (approved):

illusional:
^ This doesn't technically need a bumpversion as there's no change to the CLI, just a release to the server which is manual anyway

PR #576, opened 2023-01-12
Generate CPG config from server

illusional:
You can run `analysis-runner config` to generate an example config, can be useful for checking params without going through an analysis-runner invocation.


vladsavelyev (approved):
It looks like there is a lot of repetitiveness between the new `config` and the `index` endpoints, should the config building code be wrapped in a function that is also called by `index`?
Otherwise, awesome!

PR #579, opened 2023-01-24
subsetting script update

MattWellie:
An update to the Table/MatrixTable subsetting scripts:

- default MT subsetting behaviour is to remove monoallelic sites (reference blocks prior to densification) - shouldn't be an issue once we use the VDS format(?) - `--keep_all_ref` flag to override
  - Is this also required for the Hail Table script?
- `--keep_ref` flag is now `--keep_sample_ref` to differentiate from the above
  - Would anyone ever want to use this flag?
- option descriptions for the above flags are more verbose
- Hail Tables can now be exported straight to VCF if required (default is still HT -> HT, and MT -> MT)
- New option to export a MT as a HT 🤯 - requires testing
- Does this need two scripts at all? Surely this can be one script, with optional params applying to Matrix tables (samples, genotype subsetting) - might be a conversation for another PR
- Output path generation uses the recent changes to output_path

This would all benefit from a readme, which in turn would benefit from some discussion about how many of these options are probably never going to be used, and whether this pair of scripts should be consolidated into one. Apologies for this mad grab-bag PR 🙃


scripts/subset_hail_table.py
```diff
@@ -67,14 +66,23 @@ def main(
...
+        os.path.join(get_config()['workflow']['output_prefix'], output_root),
+        access_level='test',
     )
 
     # write the Table to a new output path
-    ht.write(f'{actual_output_path}.ht', overwrite=True)
+    if out_format in ['ht', 'both']:
+        table_path = f'{output_path}.ht'
+        ht.write(table_path, overwrite=True)
+        logging.info(f'Wrote new table to {table_path!r}')
```
lgruen, line 79: Do you need the `!r` here and below in the f-string? Isn't `table_path` just a simple string?

lgruen (approved):

violetbrina (commented):
Looks good to me. I see your comment about potentially refactoring this into a single script. I think it's a good idea, but for now if that would take too long I'm happy to approve this as-is.

scripts/subset_hail_table.py
MattWellie: AFAIK there was an update to pylint so it rejects format strings surrounded by quotes in favour of using !r. There wasn’t really any need for quotes here, just something to make the path written clearer in the logs.

KatalinaBobowik:
Looks good, @MattWellie. Just curious if you were also considering adding automatic sample subsetting (e.g., the first 20)?

MattWellie:
> Just curious if you were also considering adding automatic sample subsetting (e.g., the first 20)?

not a use case I had in mind! the use-case I had in mind was deterministic behaviour, moving a known subset of the main data into test to operate on. E.g. sneaking one family out of main to chase up with more manual methods. 

The only issue with sample sub setting is that to preserve family structures we’d need to get metamist querying involved, and that’s a substantial change. I’m not against it, just not a change I had in mind.

vivbak (approved):
I think this is great :)

scripts/subset_hail_table.py
```diff
@@ -67,14 +66,23 @@ def main(
...
         ht = ht.filter(hl.len(ht.alleles) == 2)
 
-    # create the output path; make sure we're only ever writing to test
-    actual_output_path = output_path(output_root).replace(
-        f'cpg-{get_config()["workflow"]["dataset"]}-main',
-        f'cpg-{get_config()["workflow"]["dataset"]}-test',
+    # create the output path; only ever writing to test
+    output_path = dataset_path(
+        os.path.join(get_config()['workflow']['output_prefix'], output_root),
+        access_level='test',
```
vivbak, line 72: Does this work? Linter is complaining.

vivbak: Oops just read your comment.

MattWellie:
@illusional @vivbak - I think I'd like to alter the behaviour of output_path/dataset_path a little, but that would require a small infra change (AFAIK, please advise - I'll make a more appropriate issue for this too)

- The per-dataset infra configs contain absolute paths to the main/web/tmp etc buckets under the `storage` key.
- There's some redundancy, e.g. the `tob-wgs` dataset will have the same content in `storage.default` and `storage.tob-wgs`
- The standard/full permission versions of these config files also contain a `test` category, i.e. `storage.default.test` and `storage.tob-wgs.test` have the same content.
- The test versions of the same configs don't have a test suffix - the default path key contains test paths. 

In this pretty niche situation I'd like to run with standard permissions, and write to the test bucket with as little manual path manipulation as possible so as not to undermine the effort put in to polishing the configs. I though there was now a way to do this using the cpg-utils `output/dataset_path` methods but looking closer I don't think that's true.

It looks like there was a [`test` argument planned for dataset_path](https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/hail_batch.py#L132-L133), but this isn't actually implemented. Without the `test` paths published in the test config files, it's not easy to implement this argument to switch output to a test bucket - any code expecting a test path would break when run in `test`, making copying scripts hard to test (ironic?).

Q. - Can we justify implementing a `storage.test` index in the config files across all access levels?

Context - hard path replacement from `-main` to `-test` in these scripts was working before, and I can just go back to that, no drama. I don't know any other situations where we are likely to make use of this behaviour.

illusional:
If you run in access-level test, can't it just grab default instead?
We can justify implementing storage.test inside the test configs if it makes it easier.

MattWellie:
> If you run in access-level test, can't it just grab default instead?

This should be fine, probably the easiest solution. That would make the fix:

* properly implement a test boolean argument to dataset_path (and output_path to pass on)
* if access_level is already test, use default
* if access_level is full, and the test boolean is used, use the test sub-key

illusional:
Sounds good to me! Are you in a position to action? Happy to approve.

PR #581, opened 2023-01-24
Update group cache members check

illusional:
Requires: https://github.com/populationgenomics/cpg-utils/pull/127 to be deployed 

Requires deploy of server + web server

### DIFF
--- a/web/main.py
+++ b/web/main.py
@@ -6,7 +6,7 @@ import mimetypes
 import os
 from flask import Flask, abort, request, Response
 
-from cpg_utils.cloud import read_secret, check_member_in_cached_group_members
+from cpg_utils.cloud import read_secret, is_member_in_cached_group
 import google.cloud.storage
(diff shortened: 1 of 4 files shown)


web/main.py
```diff
@@ -50,29 +50,21 @@ def handler(dataset=None, filename=None):
...
     if os.path.basename(filename) == '.access':
-        abort(403)
+        return abort(403)
 
     server_config = json.loads(read_secret(ANALYSIS_RUNNER_PROJECT_ID, 'server-config'))
-    dataset_config = server_config.get(dataset)
-    if not dataset_config:
+    if not dataset not in server_config:
         logger.warning(f'Invalid dataset "{dataset}"')
         abort(400)
```
lgruen, line 62: Should these (and potentially below) also be changed to `return`s?

lgruen (approved):

PR #590, opened 2023-03-09
Fix config option, + add traceback to server exceptions

illusional:
- Changes config from GET to POST (because it includes a config which is easier to pass in a POST body)
- Adds traceback to requests to stop us having to look at the server logs to reproduce and diagnose problems


lgruen (approved):
Perfect, thanks!

vivbak (approved):
🙌 🙏 🎉

PR #591, opened 2023-03-10
Cram releaser

EddieLF:
A script to copy crams into the release bucket of a project, based on the project name and a list of sample IDs whose crams need to be copied over.


scripts/copy_sample_cram_to_release.py
```diff
@@ -0,0 +1,65 @@
...
+def copy_to_release(project: str, path: str):
+    """
+    Copy a single file from a main bucket path to the equivalent release bucket
+    """
+    release_path = path.replace(
+        f'cpg-{project}-main',
+        f'cpg-{project}-release',
+    )
+
+    subprocess.run(f'gsutil cp {path} {release_path}', shell=True, check=True)
```
illusional, line 28: Let subprocess do the quoting for you
```suggestion
    subprocess.run(['gsutil', 'cp', path, release_path], check=True)
```

cassimons (approved):
Looks good to me

scripts/copy_sample_cram_to_release.py
```diff
@@ -0,0 +1,65 @@
...
+    logging.info(f'Copied {release_path}')
+
+
+@click.command()
+@click.option('--project', '-p', help='Metamist name of the project', required=True)
```
illusional, line 33: You can pull this from the analysis-runner config (you could leave this to provide an optional override), eg:

```
from cpg_utils.config import get_config

# this is set by the analysis-runner
config = get_config()

project = config['workflow']['dataset']
```

illusional (approved):
No issues with this, a few small nitpicks

PR #593, opened 2023-03-14
Update to copy crams into release bucket script

EddieLF:
Updated the script to use `gcloud storage cp` in favour of `gsutil cp`, and change the destination to `release/<date>` where `<date>` is todays date, e.g. `2022-03-14`, instead of the initial path to the cram replicated in release.

`gcloud storage --billing-project <project> cp <files> gs://cpg-<project>-release/<date>` 
copies multiple files in parallel into a directory in the release bucket with today's date.


MattWellie (approved):
LGTM

MattWellie:
@EddieLF  Just a thought, before running the copy would it make sense to confirm that there is a file retrieved for each sample? This job could find no CRAM files or indexes, which would then try to run a nonsensical gcloud command and fail

MattWellie (approved):

PR #594, opened 2023-03-14
once workflow finishes, parse cromwell output

MattWellie:
once a cromwell worfklow completes, read the metadata into the hail batch log


analysis_runner/cromwell.py
```diff
@@ -386,6 +387,29 @@ def _get_wait_interval(
                 continue
             status = r.json().get('status')
             _remaining_exceptions = max_sequential_exception_count
+
+            # if workflow has concluded print logging to hail batch log
+            # only do once workflow has reached a final state so as not
+            # to duplicate
+            if status.lower() in terminal_statuses:
+                logger.info('Cromwell workflow has concluded - fetching log')
+                metadata_url = (
```
illusional, line 396: This file could be enormous (like, 50MB enormous), so printing it probably isn't a good idea. I'm generally happy for the download to occur (or even better, place in the output_path called like `<workflow_id>-metadata.json`), but it would be handy to print the visualisation defined here: https://github.com/populationgenomics/analysis-runner/blob/eecb4018255c6b5db429b1b314972b5fd0fe64ab/analysis_runner/cli_cromwell.py#L289-L294

illusional (changes requested):

analysis_runner/cromwell.py
```diff
@@ -386,6 +388,32 @@ def _get_wait_interval(
                 continue
             status = r.json().get('status')
             _remaining_exceptions = max_sequential_exception_count
+
+            # if workflow has concluded print logging to hail batch log only do
+            # once workflow has reached a final state so as not to duplicate
+            if status.lower() in terminal_statuses:
+                logger.info('Cromwell workflow has concluded - fetching log')
+                metadata_url = (
```
illusional, line 396: Keen to get your thoughts, but I'd be keen for using the implementation for getting and displaying metadata from analysis_runner.cli_cromwell (or moving it to Cromwell), then you don't need to know the params of the Cromwell meta call, as you'd probably want to include subworkflows here.

MattWellie: This uses that method, indirectly - the `cromwell_cli` method renders the WorkflowMetadataModel from file, but I've set it to parse straight from the received JSON - the `WorkflowMetadataModel.display()` call is the same (line 412 ->)

MattWellie: There is the `_check_cromwell_status` method which could be dropped in instead, but the url created doesn't match this one so not sure if they're equivalent

analysis_runner/cromwell.py
```diff
@@ -386,6 +388,12 @@ def _get_wait_interval(
                 continue
             status = r.json().get('status')
             _remaining_exceptions = max_sequential_exception_count
+
+            # if workflow has concluded print logging to hail batch log
+            if status.lower() in terminal_statuses:
+                logger.info('Cromwell workflow has concluded - fetching log')
+                _check_cromwell_status(workflow_id, json_output='')
```
illusional, line 395: Or feel free to change the definition to default to None
```suggestion
                _check_cromwell_status(workflow_id, json_output=None)
```

illusional (approved):
You'll need a bumpversion commit to release this

analysis_runner/cromwell.py
MattWellie: I didn't spot this before - the type hinting is Optional but it has no default. I've set the default to `None` in the `_check_cromwell_status` method. Can you double check you're ok with that?

MattWellie: Scratch that, I see why it's there now

MattWellie:
Messy history - 

- I set the json_output to default to None, which caused a linting error
- Also ran a bump2version `minor` instead of `patch`

Reverted both and made the small suggested change plus a bump2version patch

PR #595, opened 2023-03-16
update method call and config

MattWellie:
Updated the method call, config examples, and readme

- the argument name for the method call has changed
- the config has been updated since, analysis-runner will contain (almost) all this content by default (inc. full image paths)


KatalinaBobowik:
Thanks for the update @MattWellie

KatalinaBobowik (approved):

PR #597, opened 2023-03-20
added filenames option for urls with file names

EddieLF:
This change is for when signed urls are provided with filenames as well.

e.g. in a signed_urls.txt file:

Participant_007.tar.gz https://foo.com.au/abc123?domain=garvan.org.au

Will name the file from the signed url 'Participant_007.tar.gz'


MattWellie:
The Concept works, but it will need a tweak - 

```
>>> with open('afile.txt') as handle:
...     print(handle.readlines())
...     print(handle.readlines())
...
['line1\n', 'line2\n']
[]
```

You can't run `readlines()` twice, as you already seek to the end of the file, leaving no further content to read. 

You can do `file.seek[0]` to reset the point you're reading from, or fit this into one iteration which would probably be neater

MattWellie (approved):

PR #601, opened 2023-03-24
Untar to disk

EddieLF:
This change swaps the `extractfile` method on the tar members to `extract`. 
This passes the extracted object to the disk, rather than keeping it in memory. From the disk, the file can be uploaded to the blob path in the bucket from the location it was saved to disk.


MattWellie (approved):
LGTM

PR #603, opened 2023-03-24
Download tarball, extract files, upload files, delete files, delete tarball. Simple as.

EddieLF:
Round 4....

The script has been updated to follow the following process:

1. Find tarballs in target bucket directory
2. Download a found tarball to disk and extract the contents to the disk
3. Upload and then delete each extracted file before moving on to the next
4. Delete the downloaded tarball once extracted files are uploaded
5. Go back to step 2 for the next tarball

This change also includes a check if the tarball only contains a single directory, and if it does, we go into that directory to upload those files.

The following cases should work:
- tarball contains: `<file1>`, `<file2>`, etc...
- tarball contains: `directory1/<file1>`, `directory1/<file2>`, etc.

The following cases will break the script:
- tarball contains: `directory1/<file1>`, `directory2/<file2>`
- tarball contains: `directory1<file1>`, `<file2>`

 Not sure if its worth trying to handle the breaking cases above.


scripts/untar_gz_files.py
```diff
@@ -62,25 +61,66 @@ def untar_gz_files(
...
+            logging.info(
+                f'Uploaded {file} to gs://{bucket_name}/{subdir}/{destination}/'
+            )
+
+            # Delete file after upload
+            subprocess.run(['rm', f'{filepath}'], check=True)
+            logging.info(f'Deleted {file} from disk')
+
+        # Delete tarball after all extracted files uploaded & deleted
+        subprocess.run(['rm', f'{blob_name}'], check=True)
```
cassimons, line 120: You might want to make this optional. I often would not want to rm the originals until I am 100% confident everything down stream has gone well.

cassimons (approved):

scripts/untar_gz_files.py
EddieLF: That's fair, I added this part because I'm concerned about disk space for the hail batch job. 

Since we download the tarball and extract it into disk, we might be holding up to ~180GB on disk for the biggest tarballs (90GB tarball + 90GB extracted files). I'm not sure if the hail batch job will even have enough storage capacity to complete this, even with deletions each iteration.

Also noting this wouldn't delete the tarball from the bucket, just from the hail batch disk.

cassimons:
Sorry I missed that. I thought it was deleting the copy in the bucket. All good as is then. Sent from my iPhoneOn 24 Mar 2023, at 6:15 pm, EddieLF ***@***.***> wrote:﻿
@EddieLF commented on this pull request.



In scripts/untar_gz_files.py:
> +            if is_directory:
+                filepath = f'./{subdir}/extracted/{folder}/{file}'
+            else:
+                filepath = f'./{subdir}/extracted/{file}'
+
+            output_blob.upload_from_filename(filepath)
+            logging.info(
+                f'Uploaded {file} to gs://{bucket_name}/{subdir}/{destination}/'
+            )
+
+            # Delete file after upload
+            subprocess.run(['rm', f'{filepath}'], check=True)
+            logging.info(f'Deleted {file} from disk')
+
+        # Delete tarball after all extracted files uploaded & deleted
+        subprocess.run(['rm', f'{blob_name}'], check=True)

That's fair, I added this part because I'm concerned about disk space for the hail batch job.
Since we download the tarball and extract it into disk, we might be holding up to ~180GB on disk for the biggest tarballs (90GB tarball + 90GB extracted files). I'm not sure if the hail batch job will even have enough storage capacity to complete this, even with deletions each iteration.
Also noting this wouldn't delete the tarball from the bucket, just from the hail batch disk.

PR #604, opened 2023-03-24
shuffle original script with a wrapper

MattWellie:
this is a concept PR, feel free to ignore or rewrite - 

-  The wrapper script is called as an analysis-runner job
- This parses the given directory for all the available TAR.GZ files, storing the name and size
    - The stored size is the greater of (size in GB * 2), or 30
- For each of those TAR archives:
    - create a batch job with 2 cores (?)
    - authenticate the job so that it will be able to communicate with GCP
    - assign the required amount of storage
    - create a single unpack/upload job (the core method from your previous workflow, but as a dedicated script)
    - each job may also need the `copy_common_env` method called to copy in any required environment variables... unsure
- run the collected batch jobs

Reasoning: 

- this encourages parallel unpack/uploads if there are multiple TAR files
- storage for each job is appropriate to the volume of data being unpacked
- ???
- PROFIT!

Bad things...?: 

- This assigns a high minimum storage volume, but for a small TAR file the job would also be near instant, so a negligible instance cost
- In a hypothetical situation where a directory contains a billion TAR files, this would try to spawn a billion batch jobs and that probably won't work

### DIFF
--- a/scripts/untar_gz_files.py
+++ b/scripts/untar_gz_files.py
@@ -2,143 +2,85 @@
 
 
 """
-Given a gcp directory, extract all .tar.gz files in the directory
-into output_dir defined in the call to analysis runner
+Given a single TAG.GZ, extract and upload
 """
 
 import logging
 import os
-import re
 import subprocess
 import sys
 import click
 
 # pylint: disable=E0401,E0611
-from cpg_utils.config import get_config
 from google.cloud import storage
-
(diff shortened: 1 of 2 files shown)


EddieLF:
This is really great stuff Matt, its good to see an example of creating jobs with their own resources in a batch. Thanks for sharing, I'll test this out and hopefully apply it to the main buckets soon.

PR #605, opened 2023-03-27
recursively iterate through extracted paths

EddieLF:
Unfortunately, the tarballs containing sequence data contained subdirectories, which the code could not handle. 

I've refactored the block to use Pathlib's `rglob('*')` which recursively returns everything in a directory. We then iterate through the returned paths and discard any that are a directory, leaving only file paths.

Then we take the base file names from the paths, and use these to create our output blobs in the `/subdir/extracted` folder of the bucket. We upload to each blob from the full path to the extracted file on the disk.

### DIFF
--- a/scripts/untar_gz_files.py
+++ b/scripts/untar_gz_files.py
@@ -7,6 +7,7 @@ Given a single TAG.GZ, extract and upload
 
 import logging
 import os
+import pathlib
 import subprocess
 import sys
 import click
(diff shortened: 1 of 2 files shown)


scripts/untar_gz_files.py
```diff
@@ -52,27 +53,28 @@ def main(bucket: str, subdir: str, blob_name: str, outdir: str):
...
+    extracted_from_tarball = pathlib.Path(f'./{subdir}/extracted')
+    all_extracted_paths = [
+        str(path) for path in list(extracted_from_tarball.rglob('*'))
+    ]
+
+    # Discard paths that are directories, keep paths to files
+    extracted_file_paths = []
+    for extracted_path in all_extracted_paths:
+        if not os.path.isdir(extracted_path):
+            extracted_file_paths.append(extracted_path)
```
MattWellie, line 66: Chuck all this in a conditional list comprehension

```suggestion
    extracted_file_paths = [
        str(path) for path in extracted_from_tarball.rglob('*') 
        if not path.is_dir()
    ]
```

MattWellie:
Conceptual question - it looks like we're un-tarring data we're being sent, then (potentially) ignoring some of it... Is this going to cause problems downstream?

Disclaimer - I don't know how important it is that the data we send/upload is in a single directory.

Instead of doing an un-tar, then a glob, then discarding directories, you could un-tar then do an `os.walk`

```bash
(base) ╭─mwelland@C02G74NGQ05N ~
╰─➤  tree pigs
pigs # top level dir
├── fly # file
└── when  # dir
    └── scoot  # file
```

```python
from os import walk

list(os.walk('pigs'))
>>> [('pigs', ['when'], ['fly']), ('pigs/when', [], ['scoot'])]
```

I haven't used it much, but it should break out a directory traverse by 'level', showing files and nested directories at each level. This would be useful if we do want to ensure that all files are found (instead of running a `glob('*')`, followed by a `glob('*/*')`, etc.

MattWellie (approved):

PR #611, opened 2023-03-28
Hail Batch reads input, set concurrency cap

MattWellie:
This throws a lot of stuff at the wall to see what sticks, feel free to edit/ignore

1. Uses all the available git-job setup methods from `cpg_utils.git` (i.e. the repo, hash, organisation etc. are all determined based on the _current_ commit/location)
2. When `tar.gz` files are found, store their full paths, instead of path-within-bucket
3. Pulls out the job setup method calls into a dedicated method (unchanged behaviour, just rearranged in code)
5. Reads each blob-ject into each batch job in advance (should preclude failures from copying the data into the root partition)
6. Creates an output directory for each job, which becomes the extracted data directory (e.g. instead of `./storage/extracted` which could lead to clogging the root partition. Based on [This Page](https://hail.is/docs/batch/api/batch/hailtop.batch.job.Job.html#hailtop.batch.job.Job.storage) I think all ResourceFiles are mounted at /io, which is where the requested storage volume should be mounted. i.e. this should force Hail Batch into attaching the requested storage, writing the tar file into it, and unpacking the data into it, leaving the root partition unscathed)
7. changes `input_bucket` to `bucket_client` - semantics, but this approach would no longer pull input from that bucket


scripts/unzip_wrapper.py
```diff
@@ -94,24 +129,19 @@ def main(search_path: str):
...
-        prepare_git_job(
-            job,
-            organisation='populationgenomics',
-            repo_name='analysis-runner',
-            commit=COMMIT_HASH,
-        )
+        job = set_up_batch_job(blobname, blobsize)
+
+        # read this blob into the batch
+        batch_blob = get_batch().read_input(path=blobname)
```
EddieLF, line 135: From the Hailbatch docs: 
>read_input(self, path: str) -> _resource.InputResourceFile:

        Create a new input resource file object representing a single file.

        .. warning::

            To avoid expensive egress charges, input files should be located in buckets
            that are multi-regional in the United States because Batch runs jobs in any
            US region.

I think we can safely ignore this warning because our Hail batch cluster is located in Aus, but just thought I'd flag it.

MattWellie: Hmm, I think this is a fault in the documentation... The warning is valid for most circumstances, but we host our own local instance so this is not relevant to us. AFAIK we may be the only Hail Batch deployment outside the core team in the US...

EddieLF:
> 5. Creates an output directory for each job, which becomes the extracted data directory

I'm confused at how this is being done. `job.extracted` is passed to the click `--extracted` option. How is this forcing the job to mount the storage on `/io`? 

I might be thinking too simplistically, but can we pass something like `/io/extracted` to the `--extracted` option to force the storage to mount there?

MattWellie:
I wrote a big thing that didn't really answer your question... Posted on Slack. More concise answer here (hopefully)

When you use `job.name`, Hail Batch substitutes that out for a path `/io/batch/job/name` - the usual use-case is so that Hail knows where to retrieve files from if you want to copy results out after a batch completes.

The structure of the extracted path as a String here would be `/io/batch/batch_hash/outputs/output_hash/extracted`, which provides a path in the correct location, exclusive to this job. It's impossible AFAIK to know the hashes in advance, so making Hail do it keeps each job's data segregated.

scripts/unzip_wrapper.py
illusional: MSFT also have a deployment, but we were the only for a while

MattWellie:
I don't remember what this was trying to solve

PR #613, opened 2023-04-05
Remove crams + indexes and somaliers from bucket directory

EddieLF:
This script will delete CRAMs + indexes for provided samples at the specified bucket path.

For example, if a `genome` sequence has ingested as `type: exome` and the sequence goes through alignment, a bogus CRAM is created in the `main/exome/cram/` path in the bucket. 

Context: https://github.com/populationgenomics/seqr-private/issues/73

I have to say, it seems like a data loss risk to just let this script loose in the main branch... 
Are there any additional steps or checks that should be included to mitigate risk before merging this? Or perhaps a different repo where it would be safer?

### DIFF
--- a/scripts/delete_sample_crams.py
+++ b/scripts/delete_sample_crams.py
@@ -63,6 +63,7 @@ def main(delete_path: str, somaliers: bool, samples):
     ----------
     delete_path :   a gcp path containing the crams to delete.
                         e.g. gs://cpg-project-main/exome/cram
+    somaliers   :   flag if .cram.somalier files should be deleted too
     samples     :   a list of sample ids to copy to the release bucket
     """
 
--- a/scripts/delete_sample_crams.py
+++ b/scripts/delete_sample_crams.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+
+
+"""
+Given a bucket path and sample IDs, deletes cram files in
+the path for each sample listed.
+"""
+
+import logging
+import sys
+import subprocess
+import click
+
+def check_paths_exist(paths: list[str]):
+    """
+    Checks a list of gs:// paths to see if they point to an existing blob
+    Logs the invalid paths if any are found
+    """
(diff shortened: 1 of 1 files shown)


cassimons:
@lgruen: Thoughts on this?

scripts/delete_sample_crams.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+
+    sample_ids = list(samples)
+
+    # Generate the paths to the crams and crais to be deleted
+    cram_paths = []
+    for sample in sample_ids:
+        cram_paths.append(f'{delete_path}/{sample}.cram')
+        cram_paths.append(f'{delete_path}/{sample}.cram.crai')
+        if somaliers:
```
MattWellie, line 77: --somaliers defaults to False, but I think this means we only get one shot at cleaning up the Somalier files at the same time:

- run for [samples]
- always add CRAM files
- [optionally add somalier files]
- all files need to exist to trigger the deletion

AFAIK this way we can never delete CRAMs first, and Somalier files separately.. should we default to clearing those up at the same time?

EddieLF: Yeah that's a good point to consider. 
Are there any other cram file extensions that might exist? I'm noticing some cram.md5 files in one of the buckets. Maybe I should just change it so that anything that looks like `<sample_id>.cram.extension` is included in the scope of deletions

cassimons:
I share @illusional's reservations here. Manual removal of crams should a) be an infrequent occurrence - it should basically only happen if data is withdrawn from a study or we have put it in the wrong project in error. and b) it is a unique opportunity for data loss as for most projects the cram will be our only copy of read-level data.

To me, the safest option would seem to be a "run once" data management script with hardcoded file paths for deletion that is committed to the seqr-private repo. A version of the script in this PR could be used to prepare the files for deletion list as described by @illusional.

PR #615, opened 2023-04-06
suppress infinite loop failure mode

MattWellie:
I introduced a bug previously which led to an infinitely repeating workflow: https://batch.hail.populationgenomics.org.au/batches/423486/jobs/2. Instead of failing after a finite number of attempts, this was manually killed after 17 hours 😬 

The max_sequential_exception_count is supposed to run down with each failed attempt to obtain the workflow state, In this instance this was set to 25 initially:

- the workflow had succeeded
- there was no issue with the workflow status endpoint
- because the status endpoint query was successful, **the counter was reset**
- the failure occurred when forming the status output message, so the loop began anew

This alters two behaviours: 

1. A concluded workflow will only print its status logging into Hail Batch once (boolean flag)
2. The counter will only reset if the WF status is succeeded

Keeping the counter reset in so as not to abort the watching of a succeeded workflow


illusional (approved):
You'll want to bumpversion this if you want it to be released

PR #624, opened 2023-05-04
Switch to chunked response encoding

lgruen:
This allows larger responses; see https://centrepopgen.slack.com/archives/C030X7WGFCL/p1683169178941719.

### DIFF
--- a/web/main.py
+++ b/web/main.py
@@ -100,7 +100,8 @@ def handler(dataset=None, filename=None):
     if blob is None:
         return abort(404, 'File was not found')
 
-    response = Response(stream_with_context(blob.download_as_text()))
+    # Stream the response to allow large files to be served.
+    response = Response(stream_with_context(blob.open('rt')))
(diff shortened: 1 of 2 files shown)


web/main.py
```diff
@@ -100,7 +100,7 @@ def handler(dataset=None, filename=None):
     if blob is None:
         return abort(404, 'File was not found')
 
-    response = Response(blob.download_as_string())
+    response = Response(stream_with_context(blob.download_as_text()))
```
illusional, line 103: I guess not, but not way we can stream without loading the whole file into memory first?

illusional (approved):

web/main.py
lgruen: Yeah, I thought the same, but wasn't sure whether a simple `open()` will work as expected. I'll give it a try.

lgruen: Seems to be working 👍

PR #627, opened 2023-05-17
update to allow .tar files to be unzipped

EddieLF:
Back to our favourite set of scripts 💯 

Just a small change to let this work on .tar as well as .tar.gz files

### DIFF
--- a/scripts/untar_gz_files.py
+++ b/scripts/untar_gz_files.py
@@ -44,11 +44,17 @@ def main(bucket: str, subdir: str, blob_name: str, outdir: str):
         os.makedirs(f'./{subdir}')
         os.makedirs(f'./{subdir}/extracted')
 
+    # Define the extraction options based on .tar or .tar.gz extension
+    if blob_name.endswith('.tar'):
+        options = '-xvf'
+    else:
+        options = '-xzvf'
+
     # Download and extract the tarball
     input_bucket.get_blob(blob_name).download_to_filename(blob_name)
     logging.info(f'Untaring {blob_name}')
     subprocess.run(
-        ['tar', '-xzf', f'{blob_name}', '-C', f'./{subdir}/extracted'],
+        ['tar', f'{options}', f'{blob_name}', '-C', f'./{subdir}/extracted'],
         check=True,
     )
     logging.info(f'Untared {blob_name}')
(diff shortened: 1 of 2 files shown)


scripts/untar_gz_files.py
```diff
@@ -44,11 +44,17 @@ def main(bucket: str, subdir: str, blob_name: str, outdir: str):
...
+        options = '-xvf'
+    else:
+        options = '-xzvf'
+
     # Download and extract the tarball
     input_bucket.get_blob(blob_name).download_to_filename(blob_name)
     logging.info(f'Untaring {blob_name}')
     subprocess.run(
-        ['tar', '-xzf', f'{blob_name}', '-C', f'./{subdir}/extracted'],
+        ['tar', f'{options}', f'{blob_name}', '-C', f'./{subdir}/extracted'],
```
MattWellie, line 57:
```suggestion
        ['tar', options, blob_name, '-C', f'./{subdir}/extracted'],
```

MattWellie (approved):
Just a nit, but it looks good

scripts/untar_gz_files.py
EddieLF: you're right, no need to f these variables!

jmarshall:
Where is this running, and what `tar` executable is present on that platform? These days it's a pretty unusual tar that won't just implicitly decompress if necessary, so usually `tar -xf {blob_name} …` will Just Work for uncompressed or `.gz` or `.bz2` or etc.

If you do want to keep it explicit, I guess I'd suggest something like

```python
if blob_name.endswith('.gz'):
    decompression = 'z'
elif blob_name.endswith('.bz2'):
    decompression = 'j'
else:
    decompression = ''

…
subprocess.run(
 ['tar', f'-x{decompression}f', f'{blob_name}', '-C', f'./{subdir}/extracted'],
```

to keep the `xf` in one place.

This PR also adds `-v` which will change what appears in any log file. Are there a few files in these tarballs or zillions (which might make this a bad addition!)? If you do keep the `-v` addition, it would be good to mention the reason for adding it in the commit message.

EddieLF:
@jmarshall Thanks for the notes John! This will be running in a Hail batch instance, the idea is to download the tar file from the cloud onto the Hail job's temporary disk, then extract the contents on disk before uploading the extracted files back into the cloud. 

So far the tars we've had to use this have only contained a handful of files each, but the script should be adaptable to future use cases so I'm going to remove the verbose option in case we do get something with zillions of files. I don't think we'll need to be explicit in the compression method either if `-xf` will handle all cases.

jmarshall:
Sounds good — I think all the images these will be running on will have a `tar` where the simple `-xf` will handle all cases, so the code can just be simple here. (In the unlikely event it did need the `z`, we'll find out very quickly!)

PR #629, opened 2023-05-30
added billing project argument to cram releaser script

EddieLF:
This is just a small change to the script that copies CRAMs into the release bucket, adding in an extra argument `billing_project`, for cases where the GCP billing project has a different name to the Metamist dataset.

### DIFF
--- a/scripts/copy_sample_cram_to_release.py
+++ b/scripts/copy_sample_cram_to_release.py
@@ -40,7 +40,7 @@ def check_paths_exist(paths: list[str]):
     return True
 
 
-def copy_to_release(project: str, paths: list[str]):
+def copy_to_release(project: str, billing_project: str, paths: list[str]):
     """
     Copy many files from main bucket paths to the release bucket with todays date as directory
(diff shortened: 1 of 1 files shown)


scripts/copy_sample_cram_to_release.py
```diff
@@ -40,24 +40,33 @@ def check_paths_exist(paths: list[str]):
...
 
-def copy_to_release(project: str, paths: list[str]):
+def copy_to_release(project: str, billing_project: str, paths: list[str]):
     """
     Copy many files from main bucket paths to the release bucket with todays date as directory
     """
     today = time.strftime('%Y-%m-%d')
     release_path = f'gs://cpg-{project}-release/{today}/'
 
     subprocess.run(
```
MattWellie, line 50: Not specifically related to this change, but can we wrap this in a try-except and report the command line error before the job dies to make this type of debugging easier in future?

MattWellie (approved):
LGTM

PR #630, opened 2023-05-30
Generic files to release

EddieLF:
This is a script which borrows a lot from the copy_sample_cram_to_release script. Reason we need a new script is because some collaborators aren't happy with receiving just CRAMs, they want the original fastq files. 

The idea here is that you create a .txt file containing the URLs of all the files you want to copy to the release bucket. 
Then, you upload this file into a bucket and specify its path in the argument to this script.

The txt file will then be downloaded onto the Hail disk, where it reads the URLs and executes copy commands on all the files into a directory with today's date in the release bucket.

One thing I'm unsure about is the text file encoding. I created a test .txt file in vscode and it defaulted to 'us-ascii' encoding, but I'm not sure if this is generally applicable. Is there a better practice?

### DIFF
--- a/scripts/copy_files_to_release.py
+++ b/scripts/copy_files_to_release.py
@@ -98,7 +98,7 @@ def main(project: str, billing_project: str, bucket: str, url_file: str):
     blob.download_to_filename(url_file)
 
     with open(url_file, 'r', encoding='ascii') as f:
-        paths = f.readlines()
+        paths = [line.rstrip() for line in f]
 
     # Check if all paths are valid and execute the copy commands if they are
     if check_paths_exist(paths):
--- a/scripts/copy_files_to_release.py
+++ b/scripts/copy_files_to_release.py
@@ -93,7 +93,7 @@ def main(project: str, billing_project: str, url_file: str):
     path_components = get_path_components_from_gcp_path(url_file)
 
     bucket = path_components['bucket']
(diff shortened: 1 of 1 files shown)


scripts/copy_files_to_release.py
```diff
@@ -0,0 +1,122 @@
...
+        file_name = os.path.join(path_components['suffix'], path_components['file'])
+
+    input_bucket = client.get_bucket(bucket)
+    blob = input_bucket.get_blob(file_name)
+    if not blob:
+        raise RuntimeError(f'No file found at url_file path {urls_file_path}')
+
+    # Download and read the file containing the links to copy into release
+    blob.download_to_filename(file_name)
+    with open(file_name, 'r', encoding='ascii') as f:
```
MattWellie, line 106: This will work fine, but if we know we're expecting a GCP file we can just read it using cloudpathlib.

```
from cpg_utils import to_path

with to_path(urls_file_path).open(encoding='utf-8') as f:
    paths = [line.rstrip() for line in f]
```

encoding-wise, so long as you're consistent with how it's written and read you'll be fine. I'm a fan of UTF-8 for unicode handling, but if you groom out any possible special characters you should be OK.

MattWellie (approved):

scripts/copy_files_to_release.py
EddieLF: This trims out a lot of the clutter. Many thanks @MattWellie!

PR #631, opened 2023-05-31
Collected minor fixes

jmarshall:
A few items of trivia, mostly documentation, that I noticed while looking into the dataproc issue on Monday:

* Quote anything with (potential) shell metacharacters.

    This one has an effect on the startup script: pip output goes to wherever the log goes instead of accidentally being diverted to a file named `=3.8.0`. Happily there was probably no change to the package versions actually selected in all three changed lines.

* hail_update.yaml was renamed in PR #356.

* server/Dockerfile uses `${DRIVER_IMAGE}`; I'm not sure where that's intended to be set, but in any case it needs to be transmitted to the build via `--build-arg` so this is a reminder of that.

### DIFF
--- a/dataproc/init_scripts/install_common.sh
+++ b/dataproc/init_scripts/install_common.sh
@@ -32,11 +32,11 @@ pip3 install \
     gcsfs \
     pyarrow \
     sample-metadata \
-    selenium>=3.8.0 \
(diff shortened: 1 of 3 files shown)


illusional (approved):
Nice! Always good to update this :)

PR #632, opened 2023-05-31
correction to cromwell metadata API call

MattWellie:
Before diving into some main dataset GATK-SV stuff it would be neat to try and get this call ironed out.

Current version makes the correct API route call for the Cromwell metadata, but calls the Analysis runner application.

Changes: 

- Calls the correct server instance
- Generates the correct auth token

### DIFF
--- a/analysis_runner/cli_cromwell.py
+++ b/analysis_runner/cli_cromwell.py
@@ -9,7 +9,8 @@ from typing import List, Dict, Optional
 import requests
 
 from cpg_utils.cloud import get_google_identity_token
-from analysis_runner.constants import get_server_endpoint, SERVER_ENDPOINT
+from analysis_runner.constants import get_server_endpoint, CROMWELL_URL
+from analysis_runner.cromwell import get_cromwell_oauth_token
 from analysis_runner.cromwell_model import WorkflowMetadataModel
 from analysis_runner.git import (
     get_git_default_remote,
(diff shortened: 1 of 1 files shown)


illusional (changes requested):
This doesn't work for people, only for service accounts because of OAuth reasons.

Maybe add a separate method that's something like "check_cromwell_status_direct". This is on me for not providing better documentation.

MattWellie:
> This doesn't work for people, only for service accounts because of OAuth reasons.

Instead of calling out to somewhere else I've stuck it in the main flow, using the same auth token already used by the `watch`. Is that a cruddy idea?

analysis_runner/cromwell.py
```diff
@@ -364,7 +364,9 @@ def _get_wait_interval(
     terminal_statuses = {'succeeded'} | failed_statuses
     status_reported = False
     subprocess.check_output(GCLOUD_ACTIVATE_AUTH, shell=True)
-    url = f'https://cromwell.populationgenomics.org.au/api/workflows/v1/{workflow_id}/status'
+    cromwell_api = 'https://cromwell.populationgenomics.org.au/api/workflows/v1'
```
illusional, line 367: Can you make these URLs constant at the top of the method? Eventually we might want to consider pulling them from a config.

illusional (approved):
Yep, this seems like a reasonable idea to me 👍

analysis_runner/cromwell.py
MattWellie: Updated this, the URL constant was already being imported 🙃

PR #633, opened 2023-06-01
Add script for copying reference truth

MattWellie:
Twinned with https://github.com/populationgenomics/references/pull/28

Script to copy all the validation reference data from regions scattered around in cpg-validation-test, into cpg-common-main

### DIFF
--- a/scripts/copy_validation_ref_data.sh
+++ b/scripts/copy_validation_ref_data.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+# copies the reference data used for validation to cpg-common-main
+
+# SDF files for ref genome indexing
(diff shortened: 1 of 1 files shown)


illusional (approved):
Is the`gcloud storage` tool installed by default on the driver image?
Otherwise a regular `gsutil cp` should work.

I haven't looked at the specific paths to make sure they match what you've specified in https://github.com/populationgenomics/references/pull/28. But otherwise sounds good.

MattWellie:
> Is the`gcloud storage` tool installed by default on the driver image?

Just checked and it's definitely present in the cpg_workflows image, so should also be present on the driver. I'm not sure if gsutil and gcloud are now merged into a single tool, but I remember gcloud being released as a far superior tool for file transfers..

PR #639, opened 2023-06-26
Change to mounted storage directory in unzip wrapper script

EddieLF:
Add in explicit change directory command to ensure mounted storage is used for jobs


MattWellie (approved):
I hope this solves the issues, 
sincerely,
Matt

PR #640, opened 2023-06-26
Fixed syntax in job command executing python script

EddieLF:
Correcting the command line arguments with line continuation backslashes


MattWellie (approved):
yaas

PR #642, opened 2023-06-26
Use the existing command wrapper

MattWellie:
The Cromwell watcher method is called using a source wrapper which has can be pulled straight in from cpg_utils - rewriting to remove code duplication.

This PR is stacked on the prev. minimal change, and contains CI fixes

### DIFF
--- a/analysis_runner/cromwell.py
+++ b/analysis_runner/cromwell.py
@@ -503,7 +503,7 @@ def watch_workflow_and_get_output(
             max_sequential_exception_count,
             max_poll_interval,
             exponential_decrease_seconds,
-            f'{watch_job.output_json_path}',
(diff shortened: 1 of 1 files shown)


analysis_runner/cromwell.py
```diff
@@ -487,39 +486,30 @@ def watch_workflow_and_get_output(
...
+
+    watch_job.command(
+        query_command(
+            watch_workflow,
+            watch_workflow.__name__,
+            workflow_id_file,
+            max_sequential_exception_count,
+            max_poll_interval,
+            exponential_decrease_seconds,
+            f'{watch_job.output_json_path}',
```
illusional, line 506: This is a bit weird, maybe just call str?

illusional (approved):

PR #643, opened 2023-06-27
Pass workflow ID file as a string

MattWellie:
The argument to watch_workflow needs to be cast as a String - it's taken from a prior task as a hail Batch output, which needs to be cast as a String to be passed into the watcher method. 

https://batch.hail.populationgenomics.org.au/batches/424973/jobs/2


illusional (approved):
Bumpversion required

PR #644, opened 2023-08-14
Update hail in dataproc and support multiple versions

daniaki:
Context: https://centrepopgen.slack.com/archives/C04M1G5HLM9/p1690947796216669

TLDR: Fixes a few issues in the dataproc image

1. Python 3.8 is incompatible with cpg-utils (new style type annoations) so this is updated to 3.10 by bumping Ubuntu to 22.04. cpg-utils is required to authenticating with github to clone private repos and this step was failing.
2. Adds fix to dataproc image which addresses pip depdendency resolution error in deploy.yaml when dataproc cluster is being initialised 

Also updates documentation to explain these things.

---

Michael note: I'll follow-up and update the authentication of this repo in another PR.


dataproc/Dockerfile
```diff
@@ -1,29 +1,72 @@
...
+#   WARNING: Requirement '/home/hail/hail*.whl' looks like a filename, but the file does not exist
+#   ERROR: hail*.whl is not a valid wheel filename.
+#
+# This may be due to a failure at a previous step which uses the `deploy.yaml`
+# file. This file contains info on pip dependencies and the location of the hail
+# wheel file. So, if one part of this step fails, for example the pip intall from
+# the `pip_dependencies` section of this yaml file, then the hail wheel won't be
+# copied over to the cluster resulting in the above error. However, remember to
+# check your error logs for the exact error message.
+RUN if [ "$STRIP_PIP_VERSIONS" == "true" ]; then sed -i 's/[<>=~!][^|]*//g' /usr/local/lib/python3.10/dist-packages/hailtop/hailctl/deploy.yaml; fi
```
jmarshall, line 72: There's probably mostly no matches on other lines, but you might make sure this only touches the pip_dependencies line:

```suggestion
RUN if [ "$STRIP_PIP_VERSIONS" == "true" ]; then sed -i '/pip_dependencies:/s/[<>=~!][^|]*//g' /usr/local/lib/python3.10/dist-packages/hailtop/hailctl/deploy.yaml; fi
```

README.md
```diff
@@ -95,6 +95,18 @@ analysis-runner \
 For more examples (including for running an R script and dataproc), see the
 [examples](examples) directory.
 
+### GitHub Authorisation
```
jmarshall, line 98:
```suggestion
### GitHub Authentication
```

I think so anyway… who can keep the distinction between these two straight… :smile:

dataproc/Dockerfile
```diff
@@ -1,29 +1,72 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
+
+# Give this Dockerfile the ability to checkout a specific commit hash. This is
+# useful if you want to build a specific Hail version inside the image instead
+# of the latest commit on master.
+ARG COMMIT_HASH
```
jmarshall, line 6: `HAIL_COMMIT` might be a better name, to emphasise that the repo it's talking about is hail, and it doesn't need to be a hash — e.g., could well be a tag.

What is the relationship between this and `HAIL_VERSION` in the README? I think they'd be mostly the same thing.

Am I missing something, or how would you specify these ARGs if building via the `gcloud builds submit` command shown in the README?

jmarshall (commented):
Looks good in general. I have a few (dumb) questions.

If some of what you're documenting is bugs or limitations, it'd be good to capture the bug as an issue as well. (Hence the question about the config overwriting.) However most of that comes under the rubric of “we need to revisit dataproc one day and try to reduce the number of moving parts”, and in the meantime this is a useful writeup of problems we're likely to encounter.

README.md
```diff
@@ -95,6 +95,18 @@ analysis-runner \
...
+
+If you are submitting an analysis-runner job that needs to clone a private repository owned by populationgenomics on GitHub (eg submitting a script to analysis-runner from a private repository), please make sure that your configuration file contains the following section:
+
+```toml
+[infrastructure]
+git_credentials_secret_name = '<ask_software_team_for_secret_name>'
+git_credentials_secret_project = '<ask_software_team_for_secret_project>'
+```
+
+If you are specifying multiple configuration files, please make sure that this section appears in the final right-most config to avoid these settings being overwritten.
```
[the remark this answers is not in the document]
daniaki: Configs are merged by being folded right to left, so if you set `--config=A,B,C` then the values in C will overwrite those in B and A for any duplicate keys. I believe this is by design, and I've seen it documented somewhere in `cpg-utils`

dataproc/Dockerfile
```diff
@@ -1,29 +1,72 @@
...
     apt-get install -y google-cloud-sdk && \
     apt-get clean && \
-    rm -rf /var/lib/apt/lists/* && \
-    git clone https://github.com/populationgenomics/hail.git && \
-    cd hail/hail && \
+    rm -rf /var/lib/apt/lists/*
+
+# Checkout a specific commit if supplied, to install a specific Hail version.
+RUN git clone https://github.com/populationgenomics/hail.git
+RUN if [ -n "$COMMIT_HASH" ]; then cd hail && git reset --hard ${COMMIT_HASH} && cd ..; fi
```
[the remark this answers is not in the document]
daniaki: No reason! Apart from holes in my git knowledge 😅

README.md
daniaki: The various readmes were inconsistent with `Authorisation` vs`Authorization`, so this solves that problem too 🤷

dataproc/Dockerfile
daniaki: Oh, I have completely forgotten to update the README inside `./dataproc` with instructions on how to use the new build args. `HAIL_COMMIT` sounds like a better name as it captures things that aren't just commit hashes.

As far as I can tell, `HAIL_VERSION` is only used as a docker tag, for example I tried setting `HAIL_VERSION=0.2.105` and building the image from the latest commit, hoping it would build 0.2.105 but it built the latest version that HEAD was at despite this. I will update the README.

README.md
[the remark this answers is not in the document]
jmarshall: Right, but… What I don't understand is if you have another config file also specifying `infrastructure.git_credentials_secret_project = <something_else>` then one of them isn't going to work. So presumably you'll still have a problem.

Given that this is recent mostly unused functionality, I would have been surprised if other config files people were using also contained `git_credentials_secret_project` entries. However this would be the right workaround or at least alerts users to the problem, and I guess this is what @cassimons ran into, so :+1: I guess it's the right pointer to put in the documentation :shrug:

dataproc/Dockerfile
illusional: I've renamed to HAIL_COMMIT and will prepare a config.

dataproc/README.md
```diff
@@ -1,20 +1,94 @@
...
+    * [https://centrepopgen.slack.com/archives/C030X7WGFCL/p1680587094857899?thread_ts=1679886207.438349&cid=C030X7WGFCL](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1680587094857899?thread_ts=1679886207.438349&cid=C030X7WGFCL)
+    * [https://centrepopgen.slack.com/archives/C030X7WGFCL/p1669677571568879?thread_ts=1669161375.010299&cid=C030X7WGFCL](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1669677571568879?thread_ts=1669161375.010299&cid=C030X7WGFCL)
+
+1. It's just transient (eg: [https://centrepopgen.slack.com/archives/C030X7WGFCL/p1685164583164109](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1685164583164109))
+
+1. Afailure at a previous step which uses the `deploy.yaml` file. This file contains info on pip dependencies and the location of the hail wheel file. So, if one part of this step fails, for example the pip intall from the `pip_dependencies` section of this yaml file, then the hail wheel won't be copied over to the cluster resulting in the above error. However, remember to check your error logs for the exact error message.
```
MattWellie, line 92:
```suggestion
1. A failure at a previous step which uses the `deploy.yaml` file. This file contains info on pip dependencies and the location of the hail wheel file. So, if one part of this step fails, for example the pip intall from the `pip_dependencies` section of this yaml file, then the hail wheel won't be copied over to the cluster resulting in the above error. However, remember to check your error logs for the exact error message.
```

MattWellie (approved):
Just a typo, this looks really comprehensive!

dataproc/README.md
```diff
@@ -1,20 +1,94 @@
...
+
+Note, we support specifying the hail_version, but only a very select number of versions are available (due to the init scripts not always being updated).
+
+### Rebuilding the dataproc driver image
 
 ```sh
 gcloud config set project analysis-runner
-HAIL_VERSION=0.2.110
+# grab the HAIL_VERSION from analysis_runner/dataproc.py
+HAIL_VERSION=$(grep "HAIL_VERSION = '" analysis_runner/dataproc.py | awk -F\' '{print $2}')
```
jmarshall, line 67: It still works as is, but maybe the same pattern as the other one?

```suggestion
HAIL_VERSION=$(grep "DEFAULT_HAIL_VERSION = '" analysis_runner/dataproc.py | awk -F\' '{print $2}')
```

FYI for next time: I tend to do stuff like this in one via `sed`. E.g. although dropping the quotes makes this messier:

```
sed -n "/^DEFAULT_HAIL_VERSION/s/^[^']*//p" analysis_runner/dataproc.py
```

examples/dataproc/main.py
```diff
@@ -19,10 +23,12 @@
     batch,
     max_age='1h',
     packages=['click', 'selenium'],
-    init=['gs://cpg-common-main/hail_dataproc/install_common.sh'],
-    cluster_name='My Cluster with max-age=1h',
+    init=[
+        'gs://cpg-common-main/hail_dataproc/2023-11-22-mfranklin-dev/install_common.sh'
```
jmarshall, line 27: It's just an example, but are we going to update this back to something a little less bespoke…

dataproc/init_scripts/install_common.sh
```diff
@@ -14,30 +14,12 @@ fi
 conda install -c conda-forge mamba
 # Can't update Python 3.10 without removing pinned versions first.
 rm /opt/conda/miniconda3/conda-meta/pinned
-# Remove a few packages that prevent installing Python 3.10.
-mamba remove bcolz fiona pyqt spyder
 # Finally, update Python.
 mamba install python=3.10
```
jmarshall, line 18: If we're now based on 22.04, what is the pre-existing python version?

AKA where does `conda` on line 14 come from.

It's fine as is, and I'm happy with your approach of minimal changes, but I'm just trying to convince myself that the `pip3` below does still correspond to the py3.10 being installed here…

jmarshall (approved):

dataproc/README.md
illusional: I asked copilot to write this for me 🙃

dataproc/init_scripts/install_common.sh
illusional: We'll chat about this on the call, but it comes from dataproc itself, not the dataproc driver image.

examples/dataproc/main.py
illusional: Yes absolutely!

.github/workflows/copy_dataproc_init_scripts.yaml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "dataproc/init_scripts/install_common.sh"
```
illusional, line 8:
```suggestion
```

dataproc/init_scripts/install_common.sh
illusional: Not related to this comment exactly, but it did lead us to find the image-version of the dataproc image is `2.1.2-debian11`, which comes with Python 3.10.8 by default, which means we've been able to simplify the init script to not having to update python.

PR #646, opened 2023-08-17
Corrected driver image detection

MattWellie:
If driver image isn't set explicitly in arguments to `run_cromwell_workflow_from_repo_and_get_outputs` if tries to detect from OS.getenv - this is outdated, and results in falling back to ubuntu:20.04, which doesn't have any of the required gcloud software.

https://batch.hail.populationgenomics.org.au/batches/426867/jobs/1


illusional (approved):
Will need a bumpversion commit too

PR #650, opened 2023-10-10
Create copy_gencode_to_main.sh

hopedisastro:
Copy over GENCODEv42 annotation file from `test` to `main` bucket for sc association pipeline


MattWellie:
the `-r` flag here doesn't hurt, but it doesn't help - this is a single file, but recursive copy is for directories

Are you happy with gencode version 42? 44 is available

MattWellie (approved):

PR #652, opened 2023-11-08
Mount secret value directly

illusional:
Hopefully, to avoid the issue with reading secrets.
Struggling to test because it's slow to pull the image to rebuild.

So, fire away and hope.


MattWellie (approved):
https://www.youtube.com/watch?v=vu2NK5REvWM

PR #654, opened 2023-11-23
Transition examples to use cpg_utils.get_batch

illusional:
Example: https://batch.hail.populationgenomics.org.au/batches/430260/jobs/1


MattWellie (approved):
noice

PR #657, opened 2023-12-03
Use driver image as default in https transfer jobs

EddieLF:
Specifying the driver image as the image used by all submitted HTTPS transfer jobs by the generic transfer script.


MattWellie (approved):
yup, should work as a pass-thru kwarg

PR #660, opened 2023-12-14
Extend minimum watch interval

MattWellie:
Soooo... due to the scaled timeout increase, the minimum polling period of 5 seconds is repeated a number of times. This means that when running high job counts the server is hammered into oblivion straight away, and struggles to recover.

A previous change altered the max polling interval, but the result has been that the Cromwell API still dies immediately, then the watch jobs poll for status updates and receive API errors. The polling interval between errors can be increased quite a lot, but the server can take hours to return to normal function, so this has just resulted in multiple error results, spaced really far apart.

Allowing both minimum and maximum polling intervals to be set permits the jobs running to have fine grained control over frequency (for high numbers of long running jobs, not checking until ~10mins+ after submission could be rational, and would really spread the load)


illusional (approved):
Maybe we should add some randomness to spread out the requests to avoid 140 at almost exactly 10 minutes?

MattWellie:
> Maybe we should add some randomness to spread out the requests to avoid 140 at almost exactly 10 minutes?

I've been doing a `randint()` range for each of the SG IDs to create some manual jitter. This first SG stage is really the only place where there would be multiple jobs polling at once, the rest are either once total, or once per sub-cohort

PR #662, opened 2024-01-02
Link to driver job, instead of just batch

illusional:
The analysis-runner driver batch always has one job, so let's just link to it directly.


MattWellie (approved):
Links,GeneralToMeaningful

MattWellie:
I'm really liking this change

PR #663, opened 2024-01-02
Dataproc: Force reinstall of cpg-utils

illusional:
Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1702434137655029?thread_ts=1702345700.581549&cid=C030X7WGFCL

Has already been deployed.


jmarshall (approved):
Approved, though this is an example of when I'd like to see a bit of the "WHY" explanation in the commit message or as a brief comment in _install_common.sh_.

PR #666, opened 2024-01-09
Update untar script with cpg_utils import changes

EddieLF:
Lucky number 666... 

Replaces the `cpg_workflows.batch.get_batch` import with `cpg_utils.hail_batch.get_batch` in the untar script wrapper.


MattWellie (approved):
CURSED PR

PR #668, opened 2024-01-09
Unset server-secret during deploy

illusional:
Reverts part of: https://github.com/populationgenomics/analysis-runner/pull/652

We initially set the secret as it allowed us to work around the Google Cloud incident https://status.cloud.google.com/incidents/SAKLsoUFXmYay2SvpNsq#oKGDLmoneB3YgW79P7vm in which the secret manager was not available to cloud run instances - but it was possible to mount a secret value as an environment variable.

This is causing an indeterminate delay to accessing a newly published server secret (as I think it waits for the AR server to get shut down, and then restart).

Context 1: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1704237490004689?thread_ts=1704235542.771179&cid=C030X7WGFCL
Context 2: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1704780953676709?thread_ts=1704780086.320959&cid=C030X7WGFCL


MattWellie (approved):
LovelyGoodTerrificMagnificent

EddieLF (approved):

PR #667, opened 2024-01-09
Bump + pin web versions

illusional:
Pin with `pip-compile` and install with no-deps to completely pin web versions.
(Sorry, I included a small doc change)

PR #669, opened 2024-01-09
Revert Dockerfile CMD notation change

illusional:
Introduced in #667, but I didn't know you couldn't use an environment variable in the JSON formed command, IMO, this is hadolint leading me astray.


MattWellie (approved):
how frustrating

PR #671, opened 2024-01-12
Add --storage as CLI argument

MattWellie:
Adds optional command line argument `--storage`. If specified will request attached storage for the job.


server/main.py
```diff
@@ -85,6 +85,10 @@ async def index(request):
     image = params.get('image') or DRIVER_IMAGE
     cpu = params.get('cpu', 1)
     memory = params.get('memory', '1G')
+
+    # defaults to no attached storage in Driver jobs
+    storage = params.get('cpu', None)
```
illusional, line 90:
```suggestion
    storage = params.get('storage', None)
```

server/main.py
```diff
@@ -177,6 +181,8 @@ async def index(request):
     job.image(image)
     if cpu:
         job.cpu(cpu)
+    if storage:
+        job.cpu(storage)
```
illusional, line 185:
```suggestion
        job.storage(storage)
```

illusional (approved):
I'm going to apply my suggestion feedback straight away haha, otherwise this looks good.

server/main.py
MattWellie: doiii

server/main.py
MattWellie: DOIIII

PR #673, opened 2024-03-05
Add branch name as attribute

illusional:
And add some extra things in echoing comments (like a guessed script URL)

image
image


analysis_runner/git.py
```diff
@@ -129,6 +141,59 @@ def check_if_commit_is_on_remote(commit: str) -> bool:
...
+    'main.sh'
+
+    >>> guess_script_name_from_script_argument(['./test/path/main.sh', 'arg1', 'arg2'])
+    'test/path/main.sh'
+
+    >>> guess_script_name_from_script_argument(['gcloud', 'cp' 'test'])
+    None
+
+    """
+    executables = {'python', 'python3', 'bash', 'sh'}
```
MattWellie, line 165: RScript might be a solid addition to this list?

MattWellie (approved):
this is sick, i like it

analysis_runner/git.py
illusional: Ah it was in the chatgpt suggestion, I think I removed it instead of python 🤦

PR #674, opened 2024-03-06
Update linters

illusional:
- Add ruff, mypy, shellcheck
- Update linters where required
- Update GH action versions
- Removed pylint decorators
- Add CPG ID checker


nevoodoo (approved):
Looks good from a quick scan, mostly linting and better typing annotations from what I can see. As long as the `airtable` files are def supposed to be deleted, this looks good to my eye!

PR #676, opened 2024-03-11
Untar files update

EddieLF:
Automatically moves the tar file to a `/completed` folder upon successfully extracting its contents. 

This means that if the batch has to be re-run on the same bucket path due to one or more job failures, the tarballs that were successfully extracted will no longer be picked up and queued for re-extraction.


scripts/untar_gz_files.py
```diff
@@ -76,6 +76,16 @@ def main(bucket: str, subdir: str, blob_name: str, outdir: str):
         subprocess.run(['rm', f'{filepath}'], check=True)
         logging.info(f'Deleted {file} from disk')
 
+    # Move the tarball to the "completed" directory
+    subprocess.run(
+        [
+            'gsutil',
```
MattWellie, line 82: in the spirit of modernisation can we bump this to `gcloud storage`

MattWellie (approved):

jmarshall:
Does it need to also look in the `…/completed/…` path when it's first downloading the tarball, so that it can stop cleanly if this tarball has already been done? At the moment, presumably on re-run it would throw an exception and fail.

AIUI renaming — even within the same bucket — is [not a free operation](https://cloud.google.com/storage/docs/copying-renaming-moving-objects) and is really a copy+delete. I'm probably overthinking this, but could we instead [attach some metadata](https://cloud.google.com/storage/docs/viewing-editing-metadata) saying it's been done? (I haven't used this API so don't know if this is even feasible…)

PR #678, opened 2024-03-22
Move some methods to cpg-utils

illusional:
Depends on: https://github.com/populationgenomics/cpg-utils/pull/152


nevoodoo (approved):
Looks good!

PR #681, opened 2024-04-02
Stream as bytes, not text

illusional:
I initially had a check if mime_type is text, but it actually doesn't seem to matter how it's streamed if we report the type 🤷 

Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1712037069724749 (streaming a PDF to through the PDF, a non-utf8 type)

### DIFF
--- a/web/main.py
+++ b/web/main.py
@@ -109,7 +109,7 @@ def handler(  # noqa: C901
         return abort(404, 'File was not found')
 
     # Stream the response to allow large files to be served.
-    response = Response(stream_with_context(blob.open('rt')))
+    response = Response(stream_with_context(blob.open('rb')))
     response.headers['Content-Type'] = (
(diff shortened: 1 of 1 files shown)


jmarshall (approved):
Confirmed that the internal server logs show

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x81 in position 10: invalid start byte
```

for these requests. So the problem is indeed that these binary files unsurprisingly cannot be converted as UTF-8-encoded Unicode text.

So the fix will be correct for binary files. I am nervous what it'll do for files that actually are (Unicode) text, but I assume from your comment about having a draft that checked `mime_type` that it turns out not to matter for text files and `rb` works for them too. Not too surprising I guess as this is Unix and such files will be UTF-8-encoded anyway.

So let's merge this and I'll look forward to observing text files being returned happily and not noticing the difference. :smile:

PR #683, opened 2024-04-15
Fix metamist keys for ar-integration

illusional:
Params are passed as query params, which I don't love, but it's easier to change it here than create a model to capture it on metamist.

For reference:

1. analysis-runner puts a message with meta on a pubsub queue
2. This function takes that message off the queue, and puts it into metamist

### DIFF
--- a/sample_metadata/main.py
+++ b/sample_metadata/main.py
@@ -8,6 +8,7 @@ in the Sample-Metadata database.
 import base64
 import json
 from typing import Any, Dict, Literal
+from urllib.parse import urlencode
 
 import requests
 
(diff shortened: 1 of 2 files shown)


sample_metadata/main.py
```diff
@@ -54,15 +55,15 @@ def sample_metadata(data: Dict[Literal['data'], str], unused_context: Any):
         'hail_version': hail_version,
         'batch_url': batch_url,
         'submitting_user': user,
-        'meta': metadata,  # any remaining keys
         'output_path': output_prefix,
     }
+    q = '&'.join([f'{k}={v}' for k, v in query_params.items()])
```
dancoates, line 60: I think probably better to use urllib here so that any escaping is handled. 

```python
from urllib.parse import urlencode

q = urlencode(query_params)
```

dancoates (approved):
Just one small note - all good otherwise

sample_metadata/main.py
illusional: Thanks for the suggestion, have applied 🙏

PR #692, opened 2024-07-15
Fix recursive call

EddieLF:


MattWellie (approved):
Docstring looks a little iffy now - so this is a synchronous entrypoint to an asynchronous entrypoint 🙃

EddieLF:
@MattWellie cheers, yeah I'm not sure exactly how to describe it. 

This is my intuitive understanding but honestly it might be full of holes. 

We can't `await` things in standard synchronous functions, so there's no way to call the` create_sv_analyses_async` function. So, we need to use an entrypoint function that starts an event loop with asyncio, then inside the event loop we can do our awaits and run things asynchronously. 

The entrypoint function itself `create_sv_analyses` is a standard, synchronous function, but it acts as an entrypoint to the asynchronous version of the function, `create_sv_analyses_async`. In this sense I see it as a synchronous entrypoint to an asynchronous loop.


This script could use a lot more cleaning up and improving, but it's nice to have a foundation. Going to merge this in for now so we can parse the existing cromwell analyses before they disappear, but will return to this and improve it for future runs. Cheers for the review.

PR #695, opened 2024-07-29
Support repo-less execution + refactor

illusional:
- Make repo + commit optional (with use of --skip-repo-checkout flag)
- Refactor the server into better param extraction, and then application.
- Split server/main.py into ar.py and config.py

Issue: https://cpg-populationanalysis.atlassian.net/browse/SET-92

Repo-less execution: https://batch.hail.populationgenomics.org.au/batches/474490/jobs/1


analysis_runner/cli_analysisrunner.py
```diff
@@ -277,3 +251,99 @@ def _perform_shebang_check(script: str) -> None:
...
+        script (list[str]): _description_
+
+    Raises:
+        ValueError: _description_
+        SystemExit: _description_
+        SystemExit: _description_
+
+    Returns:
+        RepositorySpecificInformation: _description_
+    """
```
nevoodoo, line 279: Should the descriptions be changed here to something else? Variables to me seem pretty self explanatory but may as well?

analysis_runner/cli_analysisrunner.py
```diff
@@ -277,3 +251,99 @@ def _perform_shebang_check(script: str) -> None:
...
+    commit: str | None,
+    branch: str | None,
+    cwd: str | None,
+    script: list[str],
+) -> RepositorySpecificInformation:
+    """Do all the repository specific stuff (like fetching repo, commit, cwd)
+
+    Args:
+        repository (str | None): _description_
+        commit (str | None): _description_
```
dancoates, line 267: I'd probably question the value of a doc comment here without extra descriptions, the types convey the same info.

server/ar.py
```diff
@@ -0,0 +1,319 @@
...
+
+    return AnalysisRunnerJobArgs(
+        dataset=dataset,
+        output=output_prefix,
+        cloud_environment=cloud_environment,
+        access_level=access_level,
+        description=params['description'],
+        config=params.get('config'),
+        # repo specific
+        repo=get_and_check_repository(params, dataset_config),
```
dancoates, line 246: Can just use the var declared above rather than calling this again

dancoates (approved):
looks good, just a couple of little things

analysis_runner/cli_analysisrunner.py
illusional: Yeah agreed, the auto docstring generator wasn't much use, and doesn't add a lot of information

PR #703, opened 2024-09-19
Pip-Audit

nevoodoo:
This adds `pip-audit` to a new `security.yaml` workflow, and bumps initial set of dependencies flagged as vulnerable.


ckrishk (approved):
Excited for security audits

PR #704, opened 2024-09-19
Configure Dependabot

nevoodoo:
This sets up a basic dependabot workflow for version upgrdes that runs every midnight, and avoids major version changes.


ckrishk (approved):
looks good

nevoodoo:
We should use Renovate instead

PR #705, opened 2024-09-23
Remove gvcf_info from VCF export scripts

MattWellie:
Explicitly removes the gvcf_info field prior to VCF export, if it exists in the data set. This field contains a Struct/dict, so Hail can't map it to a valid VCF data type.

Adding to the HT and MT export, though a HT shouldn't have any genotypes/FORMAT data

No bumpversion commit here, this is scripts-only, and won't affect any analysisrunner content


Alexander-Stuckey (approved):
Looks good to me

scripts/subset_hail_table.py
```diff
@@ -89,6 +89,11 @@ def main(
 
     # if VCF, export as a VCF as well
     if out_format in ['vcf', 'both']:
+        # remove GVCF INFO field if present - can't be exported to VCF
+        # HT shouldn't contain a FORMAT fiels, this is here as insurance
+        if 'gvcf_info' in ht.row:
```
Alexander-Stuckey, line 94: After some testing, `gvcf_info` is in `ht.entry`. You need to drop it from there.

Alexander-Stuckey (changes requested):

scripts/subset_hail_table.py
MattWellie: Ahh, doi. Yeah, the Entry of a MT maps to the FORMAT block of a VCF. HTs shouldn't have any Entries (just rows & columns), so I think this comment should have been made on the other script, and the removal shouldn't be required here. 

I'll alter this to check for gvcf_info in mt.entry, and remove the check from this script. 

AFAIK the script I was copying this from works, and checks in mt.row... maybe that wasn't working after all

Alexander-Stuckey (approved):
Looks good to me

PR #706, opened 2024-09-23
Don't add comment printing command if there are no comments to print

jmarshall:
Followup to #695, which made this check in _server/ar.py_ but neglected it in _server/cromwell.py_:

When `get_batch_comments()` returns `[]`, this invoked `job.command('')` which leads to a syntax error in the eventual job's shell script.


MattWellie (approved):
LGTM, but I'm not sure if this treats my current problem as I'm finding this error mode in a non-Cromwell run

jmarshall:
Turns out the same problem in non-Cromwell runs was only fixed recently in 3.1.0, and deploying that failed (for possibly spurious reasons). So we'll merge this and try to deploy both fixes.

PR #709, opened 2024-10-31
Add new script to zip GCS file trees and upload them to Zenodo

jmarshall:
Run via analysis-runner so it has direct access to our buckets, zips via Python's zipfile library, and uploads directly to a (pre-made) Zenodo deposit via Zenodo's API.

Context: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1730142264294839?thread_ts=1728943125.475979&cid=C018KFBCR1C

### DIFF
--- a/scripts/zip_to_zenodo.py
+++ b/scripts/zip_to_zenodo.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+
+"""
+Create zip archives of the files in the specified GCS tree and
+upload them to the specified (already created) Zenodo deposit.
+
+Typical usage:
+
+analysis-runner --dataset DATASET --description 'Upload to Zenodo' \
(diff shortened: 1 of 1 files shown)


nevoodoo (approved):

dancoates (approved):
I think this looks good, I'm not 100% sure on how it will go as far as memory consumption goes. It looks like the GC should be able to clean up between files so shouldn't be too bad as long as no individual files are too big - but my intuition of python's GC whims is not too good so will be interested to hear how it goes.

jmarshall:
@dancoates It's a good point — In this case, there's ~15,000 files going in to each zip archive, for a total of almost 400,000. I don't have much test data in a test bucket, so I guess the thing to do will be to merge it and see what happens… 🤞

PR #711, opened 2024-11-01
Correction to helper message

MattWellie:
Correction to the error message reflecting the new location of the 'repository map'

- to create the personalised error message, we need to pass dataset through as an argument, which was not previously done
  - we could use a generic `cpg-infrastructure-private/blob/main/datasets/DATASET/repositories.yaml` message instead, and preserve this interface, but AFAIK it's only called internally and all instances are updated
- to preserve line-length limits, the string is established as a template initially, then populated when needed


server/util.py
```diff
@@ -289,12 +291,17 @@ def get_and_check_script(params: dict) -> list[str]:
...
 
-def get_and_check_repository(params: dict, dataset_config: dict) -> str | None:
-    repo = params.get('repo')
-    if not repo:
+def get_and_check_repository(
+    params: dict,
+    dataset_config: dict,
+    dataset: str,
+) -> str | None:
+    if not (repo := params.get('repo')) or check_allowed_repos(
```
jmarshall, line 299: `check_allowed_repos()` is a procedure-style defn, (i.e., it has no `return` statement). So it returns None which means successful for this function… TL;DR this makes for a confusing hard-to-read `if` conditional here and IMHO it would be better to leave this as a separate `check_allowed_repos(…)` statement as per the existing code.

MattWellie: Yup, that's a wiff from me. Coming back to this I've also spotted that `get_and_check_repository` calls `check_allowed_repos`, but in the Cromwell script we were calling one then the other, so there's redundancy (and I'd changed the interface for the `check_allowed_repos` method, so that would have been a bad bug to keep)

jmarshall: Yup. Don't quite understand why the outdated `check_allowed_repos` call didn't cause a type-checking linter error…

jmarshall (approved):

PR #714, opened 2024-11-20
add optiono to subset by samples

Alexander-Stuckey:
Need to subset a Table by samples, so just adding the functionality here.


MattWellie:
Does this work? I assume it works, but this is `scripts/subset_hail_table.py` - and Hail Tables are sites-only representations of VCFs (no sample columns)

MattWellie (approved):
could be good to chuck a `describe()` at the start, I have no idea what's in this table 🙃

PR #715, opened 2024-12-16
Added a default status reporter

nevoodoo:
This adds a default `metamist` status reporter to the config, so that analysis outputs are registered correctly.


violetbrina (approved):
Nice! Looks good to me.

PR #717, opened 2025-01-13
Use Trusted Publishing to upload package to PyPI

jmarshall:
This uses an automatically minted short-lived token rather than needing a token like PYPI_API_TOKEN. See <https://docs.pypi.org/trusted-publishers/>. As per best practices, we separate the uploading into a separate job (which requires uploading/download artifacts to communicate between jobs) so that only the PyPI publishing step gets the credentials.

Also upgrade setuptools to a version that produces a PEP-625-compliant archive filename, i.e., `analysis_runner-*` rather than `analysis-runner-*`, which will prevent “Deprecation notice for recent source distribution upload to 'analysis-runner'” emails from PyPI. This step can be removed when the setup-python image eventually contains setuptools 69.3.0 or later (currently it has 65.5.0).

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -21,6 +21,10 @@ jobs:
         with:
           python-version: '3.10'
 
+      # Needed until setup-python's image contains a PEP625-compliant setuptools
+      - name: Upgrade setuptools
+        run: pip install --upgrade 'setuptools>=69.3.0'
+
       - name: Build
         run: python setup.py sdist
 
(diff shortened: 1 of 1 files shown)


jmarshall:
Before merging this, we will need to:
* Set up a corresponding trusted publisher within the analysis-runner PyPI project;
* Create a `pypi` environment in this GitHub repo, and probably restrict it to only accept deployments from the **main** branch.

jmarshall:
Trusting publisher flow set up as follows:

release.yaml / pypi

nevoodoo (approved):

jmarshall:
Merging that to **main** produced an [unexpectedly failing action run](https://github.com/populationgenomics/analysis-runner/actions/runs/12799810264/job/35686586285):

```
Uploading distributions to https://upload.pypi.org/legacy/
Uploading analysis_runner-3.2.1.tar.gz
WARNING  Error during upload. Retry with the --verbose option for more details. 
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         Only one sdist may be uploaded per release.
```

This upload error should have been ignored due to `skip-existing: true`, but perhaps it did not trigger because the existing sdist file was named `analysis-runner-3.2.1.tar.gz` (with a hyphen). We will monitor this in future executions.

PR #720, opened 2025-03-06
Skinny image templates

MattWellie:
# Goal

- Produce some minimal Docker image templates which include the CPG-Fork of Hail, to provide a lean base for pipelines to build on top of.
- Don't affect the current driver/analysis-runner images or building pathway
- Don't force anyone to adopt this, just make new optional templates

# @MattWellie's Thoughts

- CPG runs fully containerised analysis pipelines, so our Docker images and their contents are important
- We use Hail (Batch), but run a private fork of the main codebase, so we need to easily build and deploy images containing that specific version
- We don't push the Hail build from our fork to PyPi, so we are only able to access our private build of Hail by building from source, with other images inheriting that content
- Analysis-runner facilitates running of arbitrary code with ease (e.g. the default analysis-runner image contains installs of hail/pyspark, metamist, analysis-runner, GCloud CLI, etc.)

... There's nothing unusual about this so far

### Problem:

- The image we typically use as a base to build on (`analysis-runner:driver`) is 9.53GB decompressed, and is currently the only route we have for building pipeline images on our specific fork of Hail
- When we want to build specific images (e.g. to run a specific workflow), we either have to build on top of this monster image, compile hail from scratch, or use the public hail build instead. Our CI pipelines only currently implement option 1.
- Our huge images need to be loaded onto GCP servers running each Job, with pulling/startup times in the minutes instead of seconds
- Our 'template' image installs the GCloud CLI even before it installs Hail, and most of our workflows don't use or need GCloud (~1GB alone)
- The build process we currently have ([here](https://github.com/populationgenomics/analysis-runner/blob/main/driver/Dockerfile.hail#L40-L52)) doesn't pin to any specific versions. This has the benefit of pulling latest versions when rebuilt, but the downside of not knowing what you're going to get until the build completes
- The canonical driver template image is built on Python 3.10, so much of our analysis codebase is only supported at 3.10. If we had a choice of python versions to build from we would have a much easier time reacting to new releases/updates

The need to rely on this template puts us in squiffy situations, e.g. [CPG-Flow](https://github.com/populationgenomics/cpg-flow/blob/main/Dockerfile#L1):

- CPG-Flow is intended to replace cpg_workflows
- CPG-Flow experiments with building using `uv` instead of pip

but

- CPG-Flow's operational model has it's image being used as a Hail Batch Driver image, so it inherits from the base Driver
- CPG-Flow's image [includes a cpg_workflows installation](https://github.com/populationgenomics/analysis-runner/blob/main/driver/Dockerfile.hail#L45)
- UV is installed on top of the driver image, so it has an inaccessible pip-managed installation in earlier layers, then a UV installation on top (rather than instead of)

(no shade at the CPG-Flow implementation specifically, it's trying to do something progressive and cool, but due to the limitations in how we can get Hail built, it has to build on top of it's predecessor. This PR would enable CPG-Flow to cleanly build on a cpg_workflows-free base)

### Goals

- Instead of aiming for a one-size-fits-all Docker philosophy, we should make it easy for specific images to be built in as lean a way as possible
- Identify the content most foundational to our analyses (our private version of Hail, optionally Hail + GCloud), and build images with that and (nearly) nothing else
- Create a collection of true templates, the minimum image required to get people where they need to go

## UV or Pip

Because Docker files are additive (removal in later layers doesn't reduce the size of the build), it's important from the start to choose Pip or UV to build an environment in. I've added twin versions of each image and github CI workflow, building either on a UV or a Pip version of the same base image (Debian distro, python version same).

The UV version runs a `uv pip install --system` install instead of generating a virtual environment. I think that's more logical than creating a segregated virtual environment inside a container, but open to discussion ofc.

The pip version runs a standard install using the system python. 

## Multi-Stage Build Process

1. Pull a slim python image at the given python version (defaulting to 3.10)
2. First stage `basic` - Install some generic apt packages
3. New build stage from `basic`: `builder`, 
    * Installs more packages
    * pulls our Hail repo by commit/SHA/branch 
    * Hail is compiled, but not installed, leaving a Wheel
5. New build stage from `basic`: `installer`, which mounts the compiled wheel and installs Hail. Doesn't need many of the development libraries, and doesn't retain the wheel when pushed.

This concludes the first build, which would have a name like `HailBase:py3.10-hailXXX`

The second build takes the resulting image as a build argument and installs GCloud on top. This would have a separate name, e.g. `HailGcloud:py3.10-hailXXX`

These builds do a better job of scrubbing cache, reducing unnecessary installs, and splits the Hail compiling job away from the Hail installation job, which saves ~3GB alone. The Gcloud installation is still pretty heavy, but I don't feel confident enough in slimmed-down alternatives like [gcloud-lite](https://github.com/tonymet/gcloud-lite).

These images also propose allowing the Python version used as a base to be flexible, unlike our current infra which is confined to 3.10.

Nothing in this build process is ever tagged "latest", a practice we've wanted to get away from for some time

## Results

If these two workflows run as intended, there would be two image pairs:

- `HailBase` and `UvHailBase`: a minimal installation of our Hail fork, with no extras. ~2.7GB
- `HailGcloudBase` and `UvHailGcloudBase`: the Hail install plus GCloud CLI. ~3.5GB

These are still huge by docker standards, but Hail/PySpark/GCloud are all huge dependencies, so little of that footprint feels avoidable

# Feedback

I'm not attached to the names here, either for the files/directories, or the final images, but I think something like this should happen sooner rather than later.


Alexander-Stuckey:
@MattWellie I think we could probably include a `docker squash` as the final part of our container CI pipeline as well, to reduce the final size of the images (and final images aren't used in caches for other images anyway right?)

MattWellie:
Saved an extra ~200MB per final image by mounting in the Hail Wheel instead of permanently copying it in.

nevoodoo:
Hey @MattWellie this looks great, I will aim to review this at some point in the coming week!

nevoodoo (approved):
This is amazing, and quite impressive. Thank you for putting this PR together, I'm keen to see how this fares in place of what we have at the moment. I imagine it'll improve further once we figure a way to make build times faster. Can't see much wrong with your changes, but I also have a lot less experience with the building and use of these images. Over to @jmarshall :)

MattWellie:
Closing as partially-implemented, will return for UV images at some point in the future

PR #721, opened 2025-03-07
Add zip_to_zenodo.py ability to filter files for upload via a regex

jmarshall:
In previous usage of this script, we were uploading **all** files found in the bucket subdirectory tree.

Now we wish to upload only a subset of them, filtering by filename matches such as `*_foo.tsv`. This adds a regular expression-based `--pattern` option, so the pattern specified here would actually be `.*_foo\.tsv`.

Context: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1741248318810529?thread_ts=1728943125.475979&cid=C018KFBCR1C

### DIFF
--- a/scripts/zip_to_zenodo.py
+++ b/scripts/zip_to_zenodo.py
@@ -18,6 +18,7 @@ so storage should be set to the expected size of the largest archive.
 """
 
 import os
+import re
 from zipfile import ZIP_DEFLATED, ZipFile, ZipInfo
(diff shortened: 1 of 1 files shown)


milo-hyben (approved):
Looking good to me, that pip audit coming from one of the hail dependencies (orjson), not sure if we can merge without fixing it or we should pin to orjson >=3.9.15

PR #722, opened 2025-03-10
Add utility script that runs zip on files in GCS buckets

jmarshall:
Decoupling the zipping from the upload to Zenodo, as they have separate failure modes and amounts of time required.

This is a simple script for simple cases (of not too many files needing to be zipped), and will likely needed to be expanded in future. But it is sufficient for what we currently want to upload to Zenodo.

### DIFF
--- a/scripts/zip_gcs_files.sh
+++ b/scripts/zip_gcs_files.sh
@@ -12,14 +12,14 @@
 # files. The script will need storage space for the final zip archive
 # and one localised input file at a time.
 
-zip_path=$1
+zip_path="$1"
 shift
 
 cd "$BATCH_TMPDIR"
 
 zip_local=$(basename "$zip_path")
 
-for file_path
+for file_path in "$@"
 do
     file_local=$(basename "$file_path")
     gcloud storage cp -P "$file_path" "$file_local"
--- a/scripts/zip_gcs_files.sh
+++ b/scripts/zip_gcs_files.sh
@@ -1,0 +1,34 @@
+#!/bin/bash -ex
+#
+# Zips up the specified GCS files into a new GCS zip archive.
+#
+# Typical usage:
+#
+# analysis-runner --dataset DATASET --description 'Zip some files' \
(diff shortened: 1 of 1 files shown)


scripts/zip_gcs_files.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash -ex
```
jmarshall, line 1: `-x` here means that each command will be logged as it is executed, so we don't need any explicit `echo` commands for logging.

violetbrina (commented):
Pretty straight forward. Happy to approve. Just two thoughts:

- Did we want to do any kind of check on the total file sizes to zip together so people don't do crazy things with this script? Or just trust, I don't mind trust haha
- I don't remember exactly if we actually figured out a resolution to our scripts problem. Should this be in transfer-private or transfer instead?

scripts/zip_gcs_files.sh
```diff
@@ -0,0 +1,34 @@
...
+# and one localised input file at a time.
+
+zip_path=$1
+shift
+
+cd "$BATCH_TMPDIR"
+
+zip_local=$(basename "$zip_path")
+
+for file_path
```
milo-hyben, line 22: Comment here regarding where file_path is coming from would be nice :-)

jmarshall:
* At the moment, I'm just going with trust. The usual usage will be that the output is going to a tmp bucket, if that helps…

* We didn't come up with an answer to “where should scripts live”, alas. Seems like transfer-private is more about keeping a record of one-off scripts that have specific (and private) paths in them, whereas here the idea is the script is a utility and paths are command-line arguments. I put it here because it's used with analysis-runner, but 🤷

milo-hyben (approved):
Approved it, just had one small comment.

scripts/zip_gcs_files.sh
jmarshall: Fair enough. I made it explicit that it's looping over `"$@"`, i.e., the remaining command-line arguments.

PR #723, opened 2025-03-11
Add -p PREFIX option to zip_gcs_files.sh to shorten command lines

jmarshall:
It turns out batches using this script with lots of GCS file paths were failing to submit because the command line was too long. (Past a certain limit, Hail communicates the script to be run via a file in a temporary bucket, and the relevant bucket permissions were not set up for this for the initial analysis-runner batch.)

This adds a `-p PREFIX` option to factor out the common prefix of similar GCS paths, which effectively pushes this limit out a bit. (And sufficiently for the current task.)

Also show the size of the final zip archive alongside its MD5 checksum and listing.

### DIFF
--- a/scripts/zip_gcs_files.sh
+++ b/scripts/zip_gcs_files.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -ex
+#!/bin/bash -e
 #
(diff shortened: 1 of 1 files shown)


Alexander-Stuckey (approved):
Looks good to me

PR #725, opened 2025-03-11
Add new script to upload specified GCS files to Zenodo

jmarshall:
This is a stripped-down version of the existing _scripts/zip_to_zenodo.py_. Run via analysis-runner so it has direct access to our buckets, and uploads directly to a (pre-made) Zenodo deposit via Zenodo's API.

This is very similar to the existing (now deprecated) script, swapping out `zip_tree()` for the new `download_from_gcs()`, removing the command line arguments pertaining to zipping, and this time making sure to put its temporary local copy of each file to be transferred in `$BATCH_TMPDIR`.

In theory it should be possible to stream directly from the bucket to the Zenodo API (e.g. you could probably do it with `gsutil cat | curl` essentially) and avoid touching the local disk (and needing a big one), but the Google Storage API is 🤮 

Tested by uploading some GCS files to a draft record in the Zenodo sandbox.

### DIFF
--- a/scripts/upload_to_zenodo.py
+++ b/scripts/upload_to_zenodo.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python3
+
+"""
+Upload files (usually zip archives) from GCS paths to the specified
+(already created) Zenodo deposit.
+
+Typical usage:
+
+analysis-runner --dataset DATASET --description 'Upload to Zenodo' \
+    --access-level standard --output-dir unused --env ZENODO_TOKEN=TOKEN \
+    python3 scripts/zip_to_zenodo.py --deposit ID GCSFILE...
+
+The script will need storage space for one zip archive at a time,
+so storage should be set sufficient for the size of the largest archive.
+"""
+
+import os
+import tempfile
+
+import click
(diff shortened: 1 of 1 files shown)


violetbrina (changes requested):
Looks good to me. Pretty straight forward. My only thought would be to add a try/catch status check to ensure the gcs file download exists with a success and you don't end up with a partially downloaded blob locally before upload.

jmarshall:
Thanks for reviewing. The GCP documentation doesn't really spell out what happens if `download_to_filename()` fails, but it has no documented return value so there's nothing to check or look at there. Presumably it does signal failures via exceptions, and as there's no `try`/`catch` the script will explode and report the exception via the usual default stack trace.

Also the advice will be for the user to compare the md5 reported by _zip_gcs_files.sh_ with that reported by Zenodo after its accepted the file, so that'll be an end-to-end verification of the contents.

Turns out the streaming function I was looking for was `blob.open()` — duh! So my complaint is with the documentation not the API… But there is something to be said for doing an explicit download then upload so you can see where any problems are, so we'll leave this as is for now.

PR #727, opened 2025-03-17
Cromwell status parser counts fix

EddieLF:
Fixes:
- Renames `counts.tsv.gz` to `coverage_counts.tsv.gz` - the pipeline expects the latter, not sure how this has gone unnoticed until now...
- Checks through all attempts for each subworkflow, usually multiple attempts happen due to preemption. Now, we check all attempts and try to pull any successful outputs from any attempt. 
- Logs the whole thing much more nicely.

See test run: https://batch.hail.populationgenomics.org.au/batches/591767/jobs/1

### DIFF
--- a/scripts/cromwell_status_parser.py
+++ b/scripts/cromwell_status_parser.py
@@ -178,7 +178,7 @@ def copy_outputs_to_bucket(
     print(f'{dataset} :: {sg_id} :: Copying outputs summary:')
     print(f'  Destination: gs://{destination_bucket_name}/sv_evidence/\n')
     for _, output in outputs.items():
-        for _, value in output.items():
+        for value in output.values():
             if value.endswith('scramble.vcf.gz'):
                 analysis_file_sizes['scramble'] = to_path(value).stat().st_size
             elif value.endswith('wham.vcf.gz'):
--- a/scripts/cromwell_status_parser.py
+++ b/scripts/cromwell_status_parser.py
@@ -40,55 +40,129 @@ def get_workflow_metadata_from_api(workflow_id: str):
         sys.exit(1)
 
 
-def parse_workflow_status_and_outputs(json_data: dict):
+def print_parsed_workflow_summary(
+    wf_id: str,
+    sg_id: str,
+    dataset: str,
+    status_dict: dict,
+    outputs: dict,
+):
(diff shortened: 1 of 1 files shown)


scripts/cromwell_status_parser.py
```diff
@@ -40,55 +40,129 @@ def get_workflow_metadata_from_api(workflow_id: str):
...
+    status: dict = {}
+    outputs: dict[str, dict] = {}
+    subworkflow_name = subworkflow.split('.')[-1]
+    if not attempts:
+        status[subworkflow_name] = 'Not Started'
+        return status, outputs
+
+    # The attempts dict is a list of attempts, check each to recover the outputs
+    # and status of the sub-workflow. The final status is the status of the last
+    # attempt, which is the most relevant.
```
MattWellie, line 99: makes sense, so instead of explicitly checking for a success, then going to that attempt and ignoring the rest, that is implicit - if any attempt succeeds it will always be the last attempt

scripts/cromwell_status_parser.py
```diff
@@ -117,18 +193,23 @@ def copy_outputs_to_bucket(
             destination_gs_url = (
                 f'gs://{destination_bucket_name}/{destination_blob_name}'
             )
+            destination_gs_url = destination_gs_url.replace(
```
MattWellie, line 196: nice

scripts/cromwell_status_parser.py
```diff
@@ -101,6 +175,8 @@ def copy_outputs_to_bucket(
     source_bucket = storage_client.bucket(source_bucket_name)
     destination_bucket = storage_client.bucket(destination_bucket_name)
 
+    print(f'{dataset} :: {sg_id} :: Copying outputs summary:')
+    print(f'  Destination: gs://{destination_bucket_name}/sv_evidence/\n')
     for _, output in outputs.items():
         for _, value in output.items():
```
MattWellie, line 181: instead of `for _, x in y.items()` you can just drop down to `for x in y.values()`

MattWellie (approved):
LGTMatt

scripts/cromwell_status_parser.py
EddieLF, line 181:
```suggestion
        for value in output.values():
```

PR #729, opened 2025-03-26
SET-502 Improve performance of analysis runner web server when loading large files

dancoates:
[change how AR webserver serves files to improve performance](https://github.com/populationgenomics/analysis-runner/commit/f3265bbbc534d73c1628f0b0e84a32d4d8036b45)

Previously the webserver was attempting to stream the response using
stream_with_context, but this method is intended for smallish text files
to be returned line by line. This made the webserver extremely slow to
respond with even medium size files, taking nearly a minute to serve a
32mb .wasm file.

The intent of streaming the response was to allow serving of large files
by not writing them to disk or storing in memory before returning them,
but functionally this would not have worked as it would take so long to
serve the file that it wouldn't be very useful.

This change uses a more traditional method of returning a file using
flask's `send_file` method which will read the file into memory before
responding with it. This won't allow larger-than-memory files to be
served but functionally we've never needed that option anyway.

[cache expensive network calls to improve performance](https://github.com/populationgenomics/analysis-runner/commit/14fabdeacdad2a0a0755c068bf15ddd2c95cdd06)

The webserver request handler had a couple of calls, one to load the
server config secret, and one to get the membership cache that were
relatively expensive, taking between 1 and 2 seconds each. This update
caches those calls for a short period of time so that web pages that
make many requests for small files will not have to make these expensive
network calls on each request.

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -13,4 +13,4 @@ COPY main.py ./
 # we use an environment varaible to set the port,
 # so we can't use the JSON notation for the CMD
 # hadolint ignore=DL3025
-CMD exec gunicorn --bind :$PORT --workers 1 --threads 8 --timeout 0 main:app
+CMD exec gunicorn --bind :$PORT --workers 4 --threads 8 --timeout 0 main:app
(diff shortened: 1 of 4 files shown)


violetbrina (approved):
This looks really good to me. A small change for a large quality of life improvement.

PR #730, opened 2025-03-26
Change back to a streaming method for web server responses

dancoates:
PR #729 aimed to improve performance by removing the use of stream_with_context from the response handling. While this did improve performance it meant that we're now hitting the [32mb response size limit](https://cloud.google.com/run/quotas) that cloud run imposes for non-streaming responses.

This update reintroduces a streaming response, but using a different method that doesn't have the same small chunk size that was causing the performance issue previously. The docs on this are pretty minimal but this is pulling together some info from both the flask docs on streaming:
https://flask.palletsprojects.com/en/stable/patterns/streaming/

and the fast api docs on streaming a file-like binary response: https://fastapi.tiangolo.com/advanced/custom-response/#using-streamingresponse-with-file-like-objects

### DIFF
--- a/web/main.py
+++ b/web/main.py
@@ -5,13 +5,13 @@ import json
 import logging
 import mimetypes
 import os
-from typing import Optional
+from typing import Generator, Optional
 
 import google.auth.transport.requests
 import google.cloud.storage
 import google.oauth2.id_token
 from cachetools import TTLCache, cached
-from flask import Flask, abort, request, send_file
+from flask import Flask, abort, request
 
 from cpg_utils.cloud import read_secret
 from cpg_utils.membership import is_member_in_cached_group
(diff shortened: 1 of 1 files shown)


web/main.py
```diff
@@ -74,7 +74,7 @@ def has_permission(email: str, dataset: str, access_file: str | None, bucket_nam
 
 
 @app.route('/<dataset>/<path:filename>')
-def handler(
+def handler(  # noqa: C901
```
dancoates, line 77: Adding the extra iterfile function took this over an arbitrary complexity limit, to fix would need to refactor out some of the other code in the handler function, not keen to do that in this PR as it would make the diff and intent of the change unclear.

violetbrina (approved):
Thanks for the links in the description. Made this a super fast review. Conceptually the returning the generator makes sense. Hope this works!

web/main.py
violetbrina: Fine by me.

PR #735, opened 2025-07-22
Miscellaneous Dockerfile.hail etc fixes to precipitate rebuilding the driver image

jmarshall:
We need to build a new driver image containing cpg-utils 5.4.1, i.e., the corrected two-argument `image_path()` as per populationgenomics/cpg-utils#185.

As an excuse to build a new image, this PR collects some miscellaneous minor fixes. Most notably, the hail build leaves 1.3Gb of cached maven, mill, and pip dependencies in `/root/.cache`, and the GCP SDK installation leaves a script in `/tmp/tmp.RANDOMHASH`.

### DIFF
--- a/driver/Dockerfile.hail
+++ b/driver/Dockerfile.hail
@@ -3,7 +3,7 @@ FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.5
 
 ARG HAIL_SHA
 
-ENV HAIL_QUERY_BACKEND service
+ENV HAIL_QUERY_BACKEND=service
 
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \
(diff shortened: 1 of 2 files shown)


driver/Dockerfile.hail
```diff
@@ -40,13 +40,14 @@ RUN apt-get update && \
...
         gcsfs \
         hail \
         pyarrow \
         sample-metadata \
         metamist \
-        selenium>=3.8.0 \
-        statsmodels
+        'selenium>=3.8.0' \
+        statsmodels && \
+    rm -rf /root/.cache /tmp/tmp*
```
dancoates, line 53: It would be good to add a comment here describing what these cache files are and why they should be deleted.

dancoates (approved):
Looks good, just one small note

PR #752, opened 2026-04-29
Temporarily install JRE 17 so we can run mill within the Hail build

jmarshall:
Temporarily install this so `mill` will run, but remove it again afterwards so the eventual driver image is unchanged.


milo-hyben (approved):
Looking good to me!

PR #754, opened 2026-05-08
Update unzip wrapper script to unzip single path only

amyminiter:
`unzip_wrapper` in its current implementation will unzip _all_ tar balls found in a directory. This PR contains updates to this script that restricts the scope to a single gcs path. After the file path is validated and the size of the tar ball is calculated, the remainder of the script proceeds as previously except with a list of tarballs of len 1. 

This change supports team Population Genomics use cases where data are delivered and untarred sequentially, often with large delays between deliveries. 

Also includes some linting changes, as pre-commit has likely been updated since this was last pushed.

### DIFF
--- a/scripts/unzip_wrapper.py
+++ b/scripts/unzip_wrapper.py
@@ -108,7 +108,11 @@ def get_tarballs_from_path(
     required=True,
 )
 @click.option(
-    '--single-path', '-s', is_flag=True, default=False, help='Restrict unzipping to a single tar ball',
+    '--single-path',
+    '-s',
+    is_flag=True,
+    default=False,
+    help='Restrict unzipping to a single tar ball',
 )
 def main(search_path: str, single_path: bool):
     """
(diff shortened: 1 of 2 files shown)


scripts/unzip_wrapper.py
```diff
@@ -122,7 +149,7 @@ def main(search_path: str):
                 --bucket {bucket_name} \
                 --subdir {subdir} \
                 --blob_name {blobname} \
-                --outdir {output_dir}
+                --outdir {output_dir} # noqa: F821
```
amyminiter: pre-commit is still really cranky about this.

jmarshall: Is it because your change deletes the definition of `output_dir` above? 🤣

EddieLF (approved):
LGTM following test runs

amyminiter:
I've retested with our dummy tarball after making changes to the code 👍

PR #755, opened 2026-05-08
Calculate image size for single-path option

amyminiter:
Execution of the update to unzip_wrapper is (reasonably) failing as the updated implementation does not calculate image size for the `--single-path` option. Add this logic. 

Likely didn't fail for testing due to the trivial size of the dummy data. 

Link to failed batch:  https://batch.hail.populationgenomics.org.au/batches/1129320/jobs/1

### DIFF
--- a/scripts/unzip_wrapper.py
+++ b/scripts/unzip_wrapper.py
@@ -33,7 +33,7 @@ UNZIP_SCRIPT = os.path.join(os.path.dirname(__file__), 'untar_gz_files.py')
 def get_commit_hash():
     return (
         subprocess.check_output(
-            ['git', 'describe', '--always'],  # noqa: S603
+            ['git', 'describe', '--always'],
         )
         .strip()
         .decode()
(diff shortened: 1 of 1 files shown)


amyminiter:
Manual test passes: https://batch.hail.populationgenomics.org.au/batches/1129324/jobs/1

scripts/unzip_wrapper.py
```diff
@@ -133,7 +133,9 @@ def main(search_path: str, single_path: str):
 
     elif single_path:
         file_path = to_path(single_path)
-        blobsize = file_path.stat().st_size
+        raw_size_bytes = file_path.stat().st_size
+        blobsize = max([30, int((raw_size_bytes // GB) * 2.5)])
```
folded, line 137: blobsize was in bytes, now is in GiB. create_job is documented as taking a size in GiB so presumably this was very wrong before.

EddieLF (approved):

scripts/unzip_wrapper.py
amyminiter: Luckily `create_job` didn't exist in this script's previous life 😌 . 
I had a catastrophic crash when asking it to untar a blob after it was merged, so i doubt it would have gone undetected for very long 😬 🫠

PR #756, opened 2026-05-19
Set 1158 Add audience change to the metamist pub sub consumer

violetbrina:
This doesn't remove the currently hardcoded url, but enables the setting of an environment variable in the sample_metadata google cloud function that listens to the analysis-runner metadata Pub Sub submissions topic.

This will allow us to create a new consumer gcloud function that will report to the metamist development server.


scripts/vep_backend/vep_batch_backend.py
```diff
@@ -5,11 +5,11 @@
 """
 
 import click
+from cpg_workflows.jobs.vep import add_vep_jobs
```
violetbrina, line 8: No idea why this file needed to have the imports sorted. I would have assumed this linting error was passing on main. Either way, the import order is fixed here to pass the pre-commit checks.

jmarshall: It's because PR #754 recently made some fairly astonishing changes to _pyproject.toml_. We should probably undo the changes to `[tool.ruff.lint.isort.sections]` at least.

metamist/main.py
```diff
@@ -0,0 +1,107 @@
...
+    driver_image = metadata.pop('driverImage')
+    config_path = metadata.pop('configPath')
+    cwd = metadata.pop('cwd')
+    environment = metadata.pop('environment')
+    hail_version = metadata.pop('hailVersion', None)
+    batch_url = metadata.pop('batch_url', None)
+    meta = metadata.pop('meta', {}) or {}
+
+    if access_level == 'test':
+        project += '-test'
```
rameshka, line 62: Hey @violetbrina, this is for my learning. In `metamist`, is there a `test` project for each project? And this `test` project is where these analysis runner logs `etc` resides?

sample_metadata/main.py
```diff
@@ -61,29 +79,29 @@ def sample_metadata(data: Dict[Literal['data'], str], unused_context: Any):
...
+        logger.info(f'Created analysis with ID = {analysis_id}')
         return analysis_id
     except requests.exceptions.HTTPError as err:
-        print(f'Failed with response: {err.response.text}')
+        logger.error(f'Failed with response: {err.response.text}')
         raise err
 
 
-def get_identity_token() -> str:
+def get_identity_token(audience: str) -> str:
```
rameshka: Minor: It looks like this function is also declared in `metamist/main.py`

metamist/main.py
violetbrina: Yeah test access has test projects

sample_metadata/main.py
violetbrina: Ah I forgot to include the delete in my commit.

rameshka (approved):
Thanks @violetbrina. LGTM !!

PR #757, opened 2026-05-20
Design: analysis-runner tw mode (Nextflow via Tower CLI)

Alexander-Stuckey:
## What this is

A **planning PR with no code yet**, opened so the design can be reviewed before any implementation lands. A detailed implementation plan exists outside the repo; this body has the design summary that matters for reviewers.

## Goal

Add an `analysis-runner tw` mode that launches Nextflow pipelines via the Seqera Platform (Tower) CLI.

## Architecture

The existing analysis-runner server (Cloud Run Service) gains a new `/tw` endpoint. Per request it dispatches a one-shot **Cloud Run Job** named `tower-runner`:

- The runner image bundles a JRE and the `tw` binary.
- Tower API credentials (`TOWER_ACCESS_TOKEN`, `TOWER_API_ENDPOINT`) are mounted via `--set-secrets` at Cloud Run Job **deploy time** — they never touch Python code, never appear in deploy YAML, never get passed through `Popen`.
- The runner entrypoint downloads the requested Nextflow config from a private GCS mirror of a CPG-controlled config repo (using ambient Workload Identity creds — no GitHub credentials at runtime), then execs `tw launch`.
- The launch result (Tower run URL + ID) is written to `gs://cpg-tower-runs/<ar_guid>/result.json`; the server reads it after the job execution completes and returns it to the CLI caller.

This is **greenfield Cloud Run**, not Hail Batch — consistent with the broader Hail-replacement direction.

## CLI surface

```
analysis-runner tw \
  --dataset fewgenomes --output-dir 2026-05/run1 \
  --description \"trio analysis\" --access-level test \
  --pipeline nf-core/sarek --revision 3.4.0 \
  --config-repo cpg-nextflow-configs --config-ref main \
  --config-path fewgenomes.config \
  -- \
  --workspace cpg --compute-env gcp-batch-prod \
  --params-file params.yaml --profile gcp
```

- AR-side flags (before `--`) are validated and audited by analysis-runner.
- Everything after `--` is forwarded verbatim to `tw launch`. Future Tower flags work without code changes here.
- The pipeline (`--pipeline`) is a **public** repo — `tw` fetches it directly.
- The config (`--config-repo` / `--config-ref` / `--config-path`) lives in a **private** CPG repo, mirrored to GCS — AR downloads from GCS, never from GitHub at runtime.
- Users do **not** pass `launch` themselves; AR constructs the full `tw launch ...` invocation so it controls the chokepoint that injects the downloaded config path.

`--config` collides with analysis-runner's existing `--config` flag for TOML configs — the `--` fence is what makes the new mode unambiguous. The `tw` mode does not inherit analysis-runner's TOML config flag.

## Assumptions — please challenge any of these

1. New server endpoint lives in the existing `server/` Cloud Run Service. (The broader Hail-replacement work may restructure the server later; this plan adds to the current shape.)
2. GCP project: `analysis-runner`. Region: `australia-southeast1`. Matches existing infra.
3. Secret Manager secret names: `tower-api-token`, `tower-api-endpoint`.
4. GCS buckets (private, uniform bucket-level access): `cpg-nextflow-configs` (config mirror, read by runner), `cpg-tower-runs` (run results, written by runner, read by server).
5. New service account \`tower-runner@analysis-runner.iam.gserviceaccount.com` with: `secretmanager.secretAccessor` on the two secrets, `storage.objectViewer` on `cpg-nextflow-configs`, `storage.objectAdmin` on `cpg-tower-runs`.
6. Existing server SA gains `roles/run.developer` on the `tower-runner` job (to invoke), `storage.objectViewer` on `cpg-tower-runs` (to read result).
7. Mirror path contract (lives in the config repo's CI, not here): `gs://cpg-nextflow-configs/<repo>/refs/heads/<branch>/<path>`, `.../refs/tags/<tag>/<path>`, `.../commits/<sha>/<path>`.
8. Tower workspace + compute environments are pre-configured in the Tower UI and referenced by name in the post-`--` tail.
9. Server-side dispatch is **synchronous** for v1 — the request handler blocks on the runner job (typical: 30–90s). Async submit + poll can be added later if latency hurts. Can add a --wait <status> flag to tw command.
10. `tw` CLI version pinned in the Dockerfile (currently v0.31.0).

## Out-of-band ops prerequisites (NOT done in this PR's follow-up code)

Before the code can be exercised end-to-end:

- Create Secret Manager secrets `tower-api-token` and `tower-api-endpoint`.
- Create GCS buckets `cpg-nextflow-configs` and `cpg-tower-runs` with uniform bucket-level access.
- Create the runner SA with the IAM grants in assumption 5; grant the server SA the additions in assumption 6.
- Add a GitHub Actions workflow in the **config repo** that mirrors content to `cpg-nextflow-configs` per the path contract in assumption 7. **Out of scope here.**
- Provision Tower workspace(s) and compute environment(s) in the Tower UI.

## Files the implementation will touch

- **Create:** `analysis_runner/cli_tw.py`, `server/tw.py`, `driver/tower_runner_entrypoint.py`, `driver/Dockerfile.tower`, `.github/workflows/deploy_tower.yaml`, plus matching unit tests in `test/`.
- **Modify:** `analysis_runner/cli.py` (register mode), `server/main.py` (register route), `server/requirements.txt` (`google-cloud-run`), `README.md` (usage + ops checklist).

## Out of scope (explicitly)

- The GitHub Action that mirrors the private config repo to GCS — lives in the config repo, not here.
- Tower workspace / compute environment provisioning.
- Any change to the existing Hail-based flow.
- Async submit + poll pattern (synchronous is v1).

## Test plan (for the eventual implementation PR)

- [ ] CLI argparse tests: known flags parsed, `--` tail forwarded verbatim, multi-`--config-path` accumulates, `--config` after `--` does not collide.
- [ ] Server handler tests (Cloud Run client + GCS mocked): correct env-var overrides built, SHA refs route to `commits/` prefix, non-zero runner exit returns 500.
- [ ] Runner entrypoint tests (subprocess + GCS mocked): config(s) downloaded, argv built correctly (with and without `--revision`), Tower URL extracted from stdout, result.json written, secrets redacted from any captured stderr.
- [ ] Manual smoke test against a real Tower workspace + a small nf-core pipeline, with a post-run log scan confirming zero leakage of `TOWER_ACCESS_TOKEN` in server or runner logs.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MattWellie:
I like the shape of this, but there's a lot I'd push back on.

# The good

The basics here are all solid

- new `tw` AR mode
- cli/server-side/docker image file implementation all looks logical
- cloudrun job + set-secrets looks like a good way of obscuring private creds
- cloudrun in general gets us in the right direction re: moving off hail batch
- launch being controlled by AR after parameter sanitsation

Potentially iffy 

- a single api token/endpoint
  - endpoint makes sense, as we only have the one instance
  - a single API token means all RD analysts can trigger popgen jobs, for instance
  - a division of rd and popgen tokens, selectable at runtime, would help to scope pipeline access
  - we could even scope this as far as creating parallel workspaces and parallel credentials?

### miscellanea

This would be a good opportunity to dump `--output-dir` and `--description`. Not sure these have a purpose, and this is a fresh CLI so we can make fresh choices. `output-dir` used to route to `config.workflow.output_prefix`, but we don't use that any more. Description is mostly pointless, it's a required field which isn't surfaced anywhere convenient, it's only visible in the extended analysis-runner database records (e.g. https://sample-metadata.populationgenomics.org.au/analysis-runner/sandbox + expanding the workflow view).

## the rest

### config

```config .. lives in a private CPG repo, mirrored to GCS — AR downloads from GCS, never from GitHub at runtime```

I don't understand this. GitHub does perfect version control, and the best case scenario for a GCS sync is parity, but it's never going to be an improvement. It's not clear how a hard file system would implement unlimited parallel branches without becoming chaotic, especially when branches can be made _and deleted_ at will. The launcher job already requires secrets/credentials, and has to run as a service account. Attaching private-github-read access to this isn't a hurdle, meaning the exact github state should be readable.

AR's current process is to use arguments/files to create a final config file, which is then persisted. The GCS artifact is the product, not the input. That would be a simpler model.

I'd endorse two parallel options - 
- directly supplied config, matching current analysis-runner behaviour. `--config x` would upload and persist the client-side config file. This isn't a regression/relaxation, and would support the standard way we run AR, preventing the additional complexity of requiring a commit/push to a non-pipeline repo each time we change a workflow option. Especially relevant during the initial test/development phase.

- repo/ref/path arguments which are combined with a github checkout by analysis-runner, then used to pull a committed config file. This would be pretty simple to do in our current infra, and removes the indirection of needing GCS to host a file mirror.

Eventually we could enforce locks, like `--config` direct supply can only be used in test environments, or only `test` _access-level_ can use non-`main` HEADs.

I don't think the noted clash of `--config` is relevant here, as TW would have a separate client-side CLI? 

Notably I'd use this as an opportunity to move away from the current CPG layered base + default + user-supplied configs. All parameters supplied in a single master config file.

###  pipeline selection

This suggests never using Seqera's saved-workflows functionality, which is fine. A `tw launch` call can run a [saved](https://github.com/seqeralabs/tower-cli/blob/master/USAGE.md#launch-a-preconfigured-pipeline) workflow (repo + commit + entrypoint + compute env), or it can run on [unsaved workflows](https://github.com/seqeralabs/tower-cli/blob/master/USAGE.md#launch-an-unsaved-pipeline). 

Until analysis-runner is working perfectly as a nextflow launcher, we have two options:
- analysts use the interface to set off pipelines from saved workflows
- analysts use the tw client to set off unsaved workflows using pipeline/compute-end/params/revision

The first of these is much more user friendly IMO, but the Seqera interface doesn't do a great job of locking down config options unless the user has a very low permission/access level. It allows for various config overrides which could cause unpredictable runs. The latter should be encouraged as it promotes full comprehension of the run request, and allows for a full config file to be provided easily. It would also be the easiest transition to the same inputs via analysis-runner.


## Extras

Access gatekeeping

- there's some indirection between service account in AR and service account/compute creds in seqera which we haven't had to deal with before. Previously the analysis-runner dataset + access level lead to an exact service account being selected
- So long as the Tower API key has permission to start a run in the workspace we can use any of the compute environments in that workspace.
- We can gatekeep  this in basic ways by making `--compute-env <env>` something that AR inserts into the tw launch call based on dataset and access level, where the compute environments are also called `dataset_access-level` e.g. `seqr_test`.
- This will indirectly solve the 'how do we do test-mode' problem in Seqera - by cooking up separate credentials in Seqera, and controlling access to them, we can ensure that `--access-level test` can only start a run with a `dataset_test` compute environment, which will in turn only have access to test buckets/metamist projects

The workspace ID `-w` should probably also be provided by analysis-runner directly, maybe from another secret? The workspace and secrets need to be in sync to work, so no sense making one user-supplied.

AR Guid

- AR should continue to create an AR guid, and this should be used to name the workflow `tw -n <name>` as well as being passed into the workflow as a parameter or label `--labels=ARGUID,...` so it can be cascaded to the jobs/billing
- AR should also register the run invocation to [metamist](https://github.com/populationgenomics/metamist/blob/dev/models/models/analysis_runner.py), unsure how much of this model can be kept exactly. The [metamist ar overview](https://sample-metadata.populationgenomics.org.au/analysis-runner/fewgenomes) might be the only centralised non-seqera store of all run invocations
- do we need AR guid as a workflow argument? We don't incorporate it into file paths etc., so maybe we don't need it available within the workflow, so long as we can cascade labels, e.g. AR GUID -> workflow name/label -> job label for billing -> dataproc label

Strict AR launch

- not sure how much direct CLI -> TW argument passage should be permitted. There's a lot of arguments which can be provided directly (e.g. in the example `--params-file` would override the github/user supplied config parameters)
- if we permit `-- <--extra-args> ..` syntax we should at least sanitise the passed-thru parameters, and limit which ones can be transmitted. 
- Ideally IMO no arguments are transmitted this way, as it encourages looseness/provides workarounds
- We own analysis-runner, so if we need to add extra pass-thru arguments to `tw` these should be reviewed, and added to the `analysis-runner/tw` CLI as required, and validated if necessary

Alexander-Stuckey:
# Review from Mr Fable

## Summary
- The security model inverts on itself: the unvalidated verbatim `--` tail chooses workspace/compute-env under a single shared Tower token, so the "chokepoint" controls only the least-relevant input. Server must own the `(dataset, access_level) → (workspace, compute-env, work-dir)` mapping.
- The flagship CLI example is unimplementable as written: `--params-file params.yaml` after `--` references a local file that won't exist inside the Cloud Run Job.
- The synchronous-dispatch model collides with existing 60 s client timeouts, the single aiohttp worker (blocking stalls all Hail submissions), and Cloud Run Job default retries × non-idempotent `tw launch` (duplicate runs).
- Audit/provenance — the repo's defining feature — is absent: nothing publishes to Pub/Sub/metamist, and mutable refs (`--config-ref main`, floating `--revision`) leave what-actually-ran unrecoverable.

### Must-fix

1. **The verbatim `--` tail defeats the stated security model — the chokepoint doesn't control anything that matters.**
   *Where:* "CLI surface" section; assumption 8. Contrast `server/util.py:125-164` (`check_dataset_and_group`) and `server/util.py:105-115` (`check_allowed_repos`).
   In the existing system, validation has teeth because the server maps `(dataset, access_level)` to a Hail token / service account (`server/util.py:345-352`) — the access level determines what the workload can touch. In the proposed design, the things that determine what the workload can touch — `--workspace`, `--compute-env`, params, profiles — all live in the unvalidated verbatim tail, executed under a single shared `TOWER_ACCESS_TOKEN`. The PR's own example demonstrates the hole: `--access-level test` paired with `--compute-env gcp-batch-prod`. A user authorized only for `fewgenomes` test can name any workspace/compute-env the shared token can reach. AR-side validation becomes theater; the chokepoint controls only the config path — and even that is bypassable, since the tail can smuggle a second `--config` or `--pre-run`. This inverts `README.md:6` ("Only allow access to production datasets through code that has been reviewed").
   *Direction:* the server must own the mapping `(dataset, access_level) → (workspace, compute-env, work-dir, outdir)` — most naturally as an extension of the existing `server-config` secret (`server/util.py:60-72`), exactly how Hail tokens are mapped today. The tail can stay for genuinely benign flags, but workspace/compute-env/work-dir must be server-injected, and the tail denylist-checked (better: allowlist-checked) for flags that override them (`--workspace`, `--compute-env`, `--work-dir`, `--config`, `--pre-run`, `--post-run`, …). "Future Tower flags work without code changes" is precisely the property you don't want for security-relevant flags.
   *Confidence:* high on the flaw; medium on the exact `tw launch` flag inventory (verify against pinned v0.31.0).

2. **The design's own example is unimplementable: local-file flags after `--` have no file to read.**
   *Where:* "CLI surface" example (`--params-file params.yaml`); "Architecture" runner description.
   `--params-file params.yaml` is forwarded verbatim to `tw launch` running inside the Cloud Run Job container, where the user's local `params.yaml` does not exist. Same for any path-valued `tw` flag (`--config`, `--pre-run`, `--post-run`). Params files are the primary way real Nextflow runs are parameterised — this is not an edge case. Whatever fixes it (CLI uploads to GCS; params inlined into the request) is a material design decision with its own security review (unreviewed user input executed by the pipeline), audit implications, and size limits (env-var overrides are capped).
   *Direction:* add an explicit AR-side `--params-file` flag (before the `--`) that the CLI reads, the server stores under `gs://cpg-tower-runs/<ar_guid>/`, and the runner injects — making params part of the audited record. Reject path-valued flags in the tail.
   *Confidence:* high.

3. **Cloud Run Job default retries × non-idempotent `tw launch` = duplicate pipeline runs.**
   *Where:* "Architecture" (result.json handshake); assumption 9; test plan ("non-zero runner exit returns 500").
   Cloud Run Jobs default to `maxRetries=3` per task; the runner's task (launch pipeline, write result.json) is not idempotent. Unaddressed sequences: `tw launch` succeeds → result.json write fails → task fails → automatic re-run → second Tower launch (up to 4 total). Job crashes before writing result.json → server polls for a file that will never exist (the design never defines completed-but-no-result.json or stuck-execution behavior). Server/CLI timeout after a successful launch → user retries → duplicate run; there is no idempotency key anywhere.
   *Direction:* specify `--max-retries=0`; define server behavior for each terminal state; write a `launched.json` marker before `tw launch` returns so a retry/poll can distinguish "never launched" from "launched, result unknown"; pass `ar_guid` as the Tower run name so duplicates are detectable. Add these states to the test plan — it currently tests only the clean non-zero-exit path.
   *Confidence:* high.

4. **Synchronous v1 contradicts the existing clients, the server's concurrency model, and its own `--wait` suggestion.**
   *Where:* assumption 9. Contrast `analysis_runner/cli_analysisrunner.py:216-221` and `cli_cromwell.py:265-272` (both POST with `timeout=60`), `server/Dockerfile:19` (single gunicorn aiohttp worker), `.github/workflows/deploy_server.yaml:120-125` (no `--timeout`, so the 300 s Cloud Run default applies).
   Three collisions: (1) the established CLI convention is `requests.post(..., timeout=60)` — a handler blocking 30–90 s typical (plus cold-start variance; the deploy workflow itself notes 3-4 min first loads) routinely exceeds it, and the user's timeout-then-retry feeds the duplicate-run problem in #3. (2) Waiting on the job with the synchronous `google-cloud-run` client blocks the event loop and stalls every concurrent request, including Hail submissions on `/` (`server/ar.py:102`). (3) If `--wait` means `tw launch --wait` inside the runner, the job runs for the pipeline's duration and no HTTP request can outlive it — incompatible with the synchronous design. The existing modes already have the right shape: submit, return a URL immediately, audit via Pub/Sub (`server/ar.py:199-209`); `tw launch` returns the run URL on stdout within seconds — the only reason to block is the result.json round-trip the design itself introduced.
   *Direction:* make v1 "dispatch job, return 202 + ar_guid immediately, CLI polls `/tw/<ar_guid>`" (matches existing patterns, dissolves all three collisions), or explicitly specify the new CLI timeout, the Cloud Run service `--timeout`, and non-blocking waiting in the handler.
   *Confidence:* high.

5. **No audit/provenance integration — the feature that defines this repo is absent from the design.**
   *Where:* "Files the implementation will touch"; test plan. Contrast `server/ar.py:157-207` and `server/cromwell.py:158-237` (every existing route publishes `get_analysis_runner_metadata(...)` to `PUBSUB_TOPIC` → metamist; `README.md:11`).
   Nothing in the architecture, file list, or test plan publishes to Pub/Sub/metamist — result.json is a handshake artifact, not an audit log. Related provenance gaps: `--config-ref main` downloads from the mutable `refs/heads/<branch>/` mirror path (assumption 7), so what "main" contained at launch is unrecoverable (nothing resolves branch→SHA); `--revision 3.4.0` on a public repo can be a branch or moving tag, with no equivalent of `get_and_check_commit`'s "HEAD is not allowed" rule (`server/util.py:312-322`) and no per-dataset repo allowlist; all runs execute as one shared token, so Tower-side attribution shows a single identity.
   *Direction:* the `/tw` handler builds the standard metadata dict (mode='tw', pipeline, revision, resolved config SHA, params digest, Tower run URL) and publishes to `PUBSUB_TOPIC`; the runner records the resolved commit SHA of the config actually used; inject `ar_guid` into the Tower run name; for `standard`/`full`, require `--revision` to be a tag or SHA. Add audit-publication to the test plan.
   *Confidence:* high.

### Should-fix

6. **Justify the Cloud Run Job indirection at all.** `tw launch` is a thin client over Seqera's REST launch API, which accepts pipeline, revision, params text, and config text in the request body. The server could call it directly (or exec a bundled `tw`) with the token from Secret Manager — exactly how it already holds Hail tokens and the Cromwell OAuth token (`server/cromwell.py:250`). That deletes the job, the result.json handshake, the second SA, two IAM grants, one bucket, and the entire sync-blocking problem. The stated rationale (secret isolation, JRE bundling) is thin: the server already holds equivalent-power secrets, and `tw` ships as a native Linux binary — no JRE needed (also corrects assumption 10's image description). If the job exists for a stronger reason, say so; right now the complexity is unmotivated. *Confidence:* medium-high (verify configText/paramsText support in the exposed API version).

7. **IAM grants broader than needed (assumptions 5–6).** `storage.objectAdmin` on `cpg-tower-runs` → `objectCreator` suffices (runner only creates result.json). `roles/run.developer` also grants update/delete on the job (a compromised server could swap the runner image); a custom role with `run.jobs.run`, `run.jobs.runWithOverrides`, `run.executions.get` is tighter — at minimum note why developer is needed. Env-var overrides are visible to anyone with execution-viewer access and are size-limited (ties into #2). Add a lifecycle policy on `cpg-tower-runs`. *Confidence:* high on objectCreator; medium on the exact runWithOverrides role mapping.

8. **Secret-handling claims need two precision fixes.** (a) An env-var-mounted secret is in the entrypoint's process environment and inherited by the exec'd `tw` — the redaction test must cover `tw -v`/debug output and any `os.environ` dump. (b) The existing error middleware returns full tracebacks to the HTTP caller (`server/main.py:36-42, 45-49`); the design should state that runner stdout/stderr is never propagated into exceptions, only into redacted logs. *Confidence:* high.

9. **`--config-ref` ambiguity in the mirror contract (assumption 7).** Three prefixes (`refs/heads/`, `refs/tags/`, `commits/`) behind one flag; the test plan's "SHA refs route to commits/" implies a 40-hex heuristic that misroutes a branch named like a short SHA and leaves branch-vs-tag unresolved (is `v1.0` a tag or branch?). Define the syntax explicitly (require `heads/x`, `tags/x`, or full SHA; or probe in a documented order). Specify behavior when the mirror lags the repo (CI race): a just-pushed config silently serves the previous content — resolving and recording the SHA (#5) at least makes this detectable. *Confidence:* high.

10. **Test-server parity and deploy story unstated.** `deploy_server.yaml:87-118` deploys `server-test`, smoke-tests, then deploys `server`; the CLI has `--use-test-server` (`analysis_runner/util.py:92-96`). Does `server-test` get a `tower-runner-test` job, against which Tower workspace? Does `/tw` get a pre-prod smoke test? Also consider a `--dry-run` flag mirroring cromwell mode (`cli_cromwell.py:144-148`) — for a mode forwarding an opaque tail, "show me exactly what will be executed" is unusually valuable. *Confidence:* high that it's unaddressed.

11. **`--output-dir` is recorded but unenforced.** Existing modes map the output prefix to a real bucket path derived from dataset+namespace (`server/util.py:190`); in `tw` mode the pipeline's actual `outdir`/`workDir` come from the tail, so `--output-dir 2026-05/run1` may bear no relation to where data lands — breaking "link the output data with the exact invocation" (`README.md:3-8`). Folds into the #1 fix: AR should inject `workDir` and an outdir param derived from `(dataset, access_level, output_dir)`. Also unaddressed: cost attribution — billing maps to dataset via Hail billing projects today (`README.md:13`); a shared Tower compute env has no per-dataset attribution unless AR injects labels. *Confidence:* high on mechanism, medium on how much Tower labeling can fix.

12. **CLI mode-dispatch footgun amplified by a two-letter mode name.** `analysis_runner/cli.py:71-74`: an unrecognized first token falls through to the default mode with the token kept in `args` — a typo like `twl` becomes a Hail-mode submission. Pre-existing, but a terse mode name and overlapping flags make silent fall-through likelier. Cheap mitigation when registering `tw`: warn or hard-error when `args[0]` looks like a mode name but isn't one. Also specify `--` handling precisely: split `sys.argv` on the first `--` before argparse sees it (argparse's `REMAINDER` is notoriously buggy), and define behavior when the tail itself contains `--`. Extend the verbatim-tail test to cover a tail containing `--` and flags identical to AR flags (`--dataset` after the fence). *Confidence:* high.

### Why

- **A chokepoint is only as strong as the degrees of freedom it removes.** "AR constructs the full `tw launch` invocation" sounds like control, but control is measured by what the user *can't* choose, not by who concatenates the argv. Findings #1/#2/#11 are the same principle: every security- or provenance-relevant parameter (identity, compute environment, work/output location, executed config) must be derived server-side from validated inputs; pass-through is only safe for parameters whose worst case is enumerated. The existing codebase embodies this — `get_hail_token`, `check_allowed_repos`, `write_config` — which is why the verbatim tail reads as a regression rather than an extension.
- **Provenance requires immutability: record what ran, not what was requested.** `--config-ref main` is a request; the SHA the mirror served is the fact. Any audit field that can change meaning after the fact (branch names, floating tags, `:latest` images — the existing `get_and_check_image` TODO at `server/util.py:339-340` fights the same battle) is a provenance hole. Resolve-then-record at the chokepoint.
- **Don't hold an HTTP request open across a multi-step backend saga.** Synchronous dispatch couples four failure domains (CLI timeout, service timeout, event loop, job retries) that submit-and-poll decouples for free — and the repo's existing modes already use that shape. When a new design is *less* async than the system it joins, interrogate that before v1, not after "if latency hurts".
- **Retries and idempotency are a package deal.** Any platform-level retry (Cloud Run Jobs' `maxRetries`, requests-level retry, user retry-on-timeout) applied to a non-idempotent action manufactures duplicates. Either make the action idempotent (launch keyed by ar_guid, detect-and-return-existing) or turn retries off and define every terminal state explicitly.
- **Done well, worth keeping:** the numbered-assumptions section is exactly how a design PR should invite review; deploy-time secret mounting is the right instinct; the GCS mirror genuinely removes GitHub credentials from the runtime path; the `--` fence is the correct answer to the `--config` collision; and a test plan that includes a post-run secret-leak log scan is rarer than it should be. The fixes above mostly *redistribute* control toward patterns this codebase already has — they don't require new invention.

PR #758, opened 2026-07-05
Client side tests for analysis-runner

Alexander-Stuckey:
Adding some client side tests for analysis-runner, before I make any changes to include Nextflow tower as a supported dispatch mode.

__ALSO__ centralise dependency management into `pyproject.toml`

- Add tests for both `cli.py` and `cli_analysisrunner.py`.
- Minimal tests to ensure that jobs route correctly to cromwell.
- Bump version of `mypy` and `ruff`, as they were quite outdated.
- Centralise dependencies into `pyproject.toml`, delete `setup.py` etc, update build references.


MattWellie (approved):
Good start. You can push the pre-commit versions even higher - pre-commit in general is at 6.0.0+, mypy has a version v2.1.0.

The `get_repository_specific_information` method is probably running as part of this call, and will pass if _this_ testing branch is currently pushed to github which is a bit meta. Probably worth adding those checks in, as it's something we'll be re-using in TW... unless we exclusively stick with named workflows, I can't remember if we came to a decision on this.

MattWellie:
Confirming a few things from playing locally - 
1. the dev dependencies aren't present, ideally `requirements-dev.in` should have pytest and pytest-mock listed
2. the dev dependencies are kinda... distributed at the moment in a couple of requirements files. Might be a nice upgrade to roll these into a `extras_require` section in the [setup.py](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies). The more modern approach would be to roll all the requirements, setup.py, and pyproject.toml into just a pyproject.toml... easier to manage too

in the tests:
1. the [git checks](https://github.com/populationgenomics/analysis-runner/blob/main/analysis_runner/cli_analysisrunner.py#L257) are present and un-mocked, this is doing a real check against the local and remote git history, and will fail if an un-pushed commit it added locally - failing stdout example
```
The commit "fe7d23eab841cf77619ec6c05270f2a82489d012" was not found on the remote (Github). 
Please confirm if you want to proceed anyway. (y/n): 
```

--

Feels bad raising these as issues and just... passing them over, lmk if you want to combine on this

MattWellie (approved):
LGTMatt - doesn't touch functional code, adds tests, updates/modernises requirements and packaging. What's not to love? Maybe a version bump and a second approval from software 👀

test/client/test_cli.py
```diff
@@ -0,0 +1,111 @@
...
+def test_no_specified_mode_submits_analysisrunner(mocker) -> None:  # noqa: ANN001
+    mock_run_analysis_runner_from_args = mocker.patch(
+        'analysis_runner.cli.run_analysis_runner_from_args',
+    )
+    mock_run_cromwell_from_args = mocker.patch(
+        'analysis_runner.cli.run_cromwell_from_args',
+    )
+    mock_run_config_from_args = mocker.patch(
+        'analysis_runner.cli.run_config_from_args',
+    )
```
violetbrina: I'd probably make this a fixture given it's used in every test, just to reduce the duplication a little bit.

.gitignore
```diff
@@ -131,3 +131,4 @@ dmypy.json
 .idea/
 .DS_Store
 .vscode/
+tmp_review_feedback.md
```
violetbrina: Probably don't need this in the git ignore if it's a temporary artifact? I assume we'd just not commit it in the first place.

test/client/test_cli_analysisrunner.py
```diff
@@ -0,0 +1,120 @@
...
+        'access_level': 'full',
+        'script': ['echo', 'hello world'],
+    }
+    caplog.set_level(logging.INFO, logger='analysis_runner')
+    monkeypatch.setattr(
+        'builtins.input',
+        lambda prompt: 'y',  # noqa: ARG005
+    )
+    run_analysis_runner(**args)
+    assert len(posted_requests) == 4  # noqa:PLR2004  # headers, json, timeout, url
```
violetbrina: I think I'd prefer an actual key or name check here rather than length which is a more fragile test.

test/client/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+    monkeypatch.setattr('requests.post', _post)
+    monkeypatch.setattr(
+        'analysis_runner.cli_analysisrunner.get_google_identity_token',
+        lambda endpoint: 'fake-token',  # noqa: ARG005
+    )
+    return calls
+
+
+def _perform_version_check() -> None:
+    pass
```
violetbrina: Only runs at import time here, it would be better to use the autouse=True pytest fixture. See _no_outbound_requests.

pyproject.toml
```diff
@@ -1,3 +1,63 @@
...
+test = [
+    "pytest>=9.0.2",
+    "pytest-mock>=3.15.1",
+]
+dev = [
+    "analysis-runner[test]",
+    "bump2version",
+    "pre-commit",
+    "pylint",
+    "ruff~=0.11",
```
violetbrina: I'm sure you just moved the existing requirements over which is great. But if we've upgraded to ruff 0.15 in the pre-commit, we should do the same in the dev dependencies as well.

Also, we should consider adding mypy here as well and also adding the same version in pre-commit. Just to prevent situations where local runs of ruff/mypy are different from the version the pre-commit tool uses.

violetbrina (changes requested):
Thanks so much for your work adding tests to the analysis-runner, a much overdue addition for sure. I have some very minor comments, and I'm not sure if @dancoates or anyone else from @populationgenomics/software-team wants eyes on this as well.

But looking good.

.gitignore
Alexander-Stuckey: Yeah I fixed this up in the other PR (I believe), but forgot to clean this up here.

Alexander-Stuckey:
@violetbrina , I believe I've addressed everything, thanks for the feedback.

requirements-dev.txt
```diff
@@ -2,59 +2,123 @@
...
 #
+aiodns==2.0.0
+    # via hail
+aiohappyeyeballs==2.7.1
+    # via aiohttp
+aiohttp==3.14.1
+    # via hail
+aiosignal==1.4.0
+    # via aiohttp
+analysis-runner[test] @ file:///opt/deps
```
MattWellie:
```suggestion
analysis-runner[test]
```

This is breaking CI

MattWellie: I've made a side PR stacked on this branch - it un-nests the analysis-runner [dev] and [test] optional dependencies, and changes how the package is installed in CI.

It also rebuilds the requirements-dev file with uv instead of pip - it should be equivalent, but the file header is explicit about being created with UV Which we may not want

https://github.com/populationgenomics/analysis-runner/pull/760

Alexander-Stuckey:
@nevoodoo This is a package of tests that I wrote for the client side of analysis-runner. I had some packaging changes here as well, but in light of the work you are doing, have taken them out.

PR #759, opened 2026-07-07
Centralise deployment requirements

Alexander-Stuckey:
PR https://github.com/populationgenomics/analysis-runner/pull/758 started centralising project dependencies of analysis runner into `pyproject.toml`, as I wanted to do that when I implemented the tests for the client side.

This PR also moves the rest of the dependencies into `pyproject.toml`, which means that:
- server
- web
- metamist
are all managed by `[dependency-groups]` in `pyproject.toml`.

I've attempted to keep all versions of packages the same, but some will have been upgraded where there wasn't an explicit pin.

The `requirements.txt` files are now generated automatically in the `Makefile` via `uv pip-compile`

Python versions have stayed at 3.10 for server and web, vs 3,11 for Metamist. Not opposed to bumping all to 3.11, but just wanted to start from what was already in the repo.


MattWellie (commented):
I haven't used dependency groups before, are they in addition to the main project dependencies or instead of? If it's in addition to, some of these could be slimmed down further and have the same effect, but I'll have to read up on how they work

jmarshall:
They're [entirely separate](https://packaging.python.org/en/latest/specifications/dependency-groups/) from build and project dependencies. One day you might be able to include build or project dependencies from a dependency group, but they left anything like that to a future PEP.

violetbrina:
Other than it not passing the current ruff checks (which should be easy enough to handle) it looks good to me. Maybe we should have a CONTRIBUTING.md file like in cpg-flow with clear instructions on dependency management (especially with the now multiple dependency groups) and using uv for compiling requirements files and bumping.

I'd also lean towards a full use of uv in the repo and not just for compiling requirements. There's no reason we cannot update the analysis-runner to have it's version managed by uv and having the uv.lock file for the main dependencies. A setup like in cpg-flow is a good example if you need one to base it off of. 

`uv pip compile` is also a drop in replacement for the pip compile if you want to keep the separate requirements docs, but a `uv sync --all-groups` is a nice way to setup your dev environment with all of the requirements. But even for the non-PEP 735 groups (e.g. [project.optional-dependencies]) you can install them with `uv sync --all-extras`. 

In that scenario I'd document the usage with uv in a contributing file and just scrap the various compiled requirements files anyway. But if people prefer we can still keep them and just use `uv pip compile` and `uv pip sync server/requirements.txt` for example.

Alexander-Stuckey:
> Other than it not passing the current ruff checks (which should be easy enough to handle) it looks good to me. Maybe we should have a CONTRIBUTING.md file like in cpg-flow with clear instructions on dependency management (especially with the now multiple dependency groups) and using uv for compiling requirements files and bumping.
> 
> I'd also lean towards a full use of uv in the repo and not just for compiling requirements. There's no reason we cannot update the analysis-runner to have it's version managed by uv and having the uv.lock file for the main dependencies. A setup like in cpg-flow is a good example if you need one to base it off of.
> 
> `uv pip compile` is also a drop in replacement for the pip compile if you want to keep the separate requirements docs, but a `uv sync --all-groups` is a nice way to setup your dev environment with all of the requirements. But even for the non-PEP 735 groups (e.g. [project.optional-dependencies]) you can install them with `uv sync --all-extras`.
> 
> In that scenario I'd document the usage with uv in a contributing file and just scrap the various compiled requirements files anyway. But if people prefer we can still keep them and just use `uv pip compile` and `uv pip sync server/requirements.txt` for example.

I deliberately left the linting broken at the moment, as that is all in actual prod code, and I wanted to update `pyproject` and write tests first. So far I've only written tests for the client side of analysis runner, and I haven't looked at the server side. Only after I'd written those tests would I want to fix all the lint errors and actually touch prod code, as I'm confident that I'll break something by mistake if I don't have the tests in place.

I'll take a look at cpg-flow for the full UV implementation, thanks 😄

violetbrina (approved):

dancoates:
I'm actually going to close this one as we want to go in a different direction with dep management, thanks for putting together @Alexander-Stuckey though. I have been a bit capacity blocked on the other repo layout change but have resource available to get it done in the upcoming software sprint starting today.

PR #760, opened 2026-07-15
alt build

MattWellie:
Fixes requirements-dev.txt, and avoids using it in CI build

Does not fix all the linting errors, but does prevent the linting workflow from throwing setup errors


.github/workflows/lint.yaml
```diff
@@ -18,8 +18,7 @@ jobs:
 
     - name: Install packages
       run: |
-        pip install .
-        pip install -r requirements-dev.txt
+        pip install ".[dev,test]"
```
Alexander-Stuckey, line 21: @violetbrina is this where we could use `uv sync --extra dev --extra test`?

Alexander-Stuckey (commented):
Just one comment on something that I'm not familiar with

.github/workflows/lint.yaml
MattWellie: I believe so. The current uv.lock (not committed) looks like 
```
...
[package.optional-dependencies]
dev = [
    { name = "bump2version" },
    { name = "mypy" },
    { name = "pre-commit" },
    { name = "pylint" },
    { name = "ruff" },
    { name = "toml" },
]
test = [
    { name = "pytest" },
    { name = "pytest-mock" },
]
...
```

and that uv sync command works as expected

analysis_runner/dataproc.py
```diff
@@ -5,20 +5,20 @@
 
 @deprecated(reason='Use cpg_utils.dataproc.setup_dataproc instead')
 def setup_dataproc(*args, **kwargs):  # noqa: ANN002, ANN003
-    from cpg_utils.dataproc import setup_dataproc as _setup_dataproc
+    from cpg_utils.dataproc import setup_dataproc as _setup_dataproc  # noqa: PLC0415
```
MattWellie, line 8: I would be very ok with removing all these deprecated methods - we don't use them, so we don't need to soft fail

MattWellie:
Probably irrelevant WRT #761

PR #762, opened 2026-07-29
SET-1225 update analysis runner metamist integration to use metamist package

hrrsheen:


metamist/main.py
```diff
@@ -39,70 +40,37 @@ def main(data: dict[Literal['data'], str], unused_context: Any):
...
+        'batch_url': metadata.get('batch_url', ''),
+        'submitting_user': metadata.get('user', ''),
+        'output_path': metadata.get('output', ''),
+        'request_body': metadata.get('meta', {}),
+        'cwd': metadata.get('cwd'),
+        'hail_version': metadata.get('hailVersion'),
     }
-    q = urlencode(query_params)
+
+    api_client = ApiClient(Configuration(host=audience))
```
dancoates: The need to pass the audience is also outdated, the metamist client handles this now. If the metamist URL needs configuring we can set the `SM_URL` env variable https://github.com/populationgenomics/metamist/blob/dev/openapi-templates/configuration.mustache#L25

so you don't actually need to set up the API client, the defaults will work fine, ie. the line below can be `ar_api = AnalysisRunnerApi()` without passing in a client.

metamist/main.py
```diff
@@ -39,70 +40,37 @@ def main(data: dict[Literal['data'], str], unused_context: Any):
...
-        'description': description,
-        'driver_image': driver_image,
-        'config_path': config_path,
-        'cwd': cwd,
-        'environment': environment,
-        'hail_version': hail_version,
-        'batch_url': batch_url,
-        'submitting_user': user,
-        'output_path': output_prefix,
+        'repository': metadata.get('repo', ''),
```
dancoates: There's a bit of a gotcha here - the metadata dict comes from a json value that is constructed here:
https://github.com/populationgenomics/analysis-runner/blob/main/server/util.py#L192

This will have all keys present with json values as `null` when there is no value. This then gets loaded at line 34 above where it will become python `None`s. With the old code, when these were run through `urlencode` they became a string of `"None"`. 

Metamist enforces that fields like repository and commit must have a value, this worked with the old code because it accepted `"None"` as the repository and commit value.

Image

But with this change it will start causing errors because the value will be `None` (ie, not a string) which isn't accepted. The fallback to an empty string in the `.get` method won't work either because the keys will always exist. 

The old code is kinda buggy but I think we need to be consistent with the old functionality so that there isn't a change to how these are logged.

So I think the best approach would be to wrap the values in `str()` which is what urlencode was doing. 

ie. 

```python
kwargs = {
        'project': project,
        'ar_guid': str(metadata.get('ar-guid')),
        'access_level': str(access_level),
        'repository': str(metadata.get('repo')),
        'commit': str(metadata.get('commit')),
        'script': str(metadata.get('script')),
        'description': str(metadata.get('description')),
        'driver_image': str(metadata.get('driverImage')),
        'config_path': str(metadata.get('configPath')),
        'environment': str(metadata.get('environment')),
        'batch_url': str(metadata.get('batch_url')),
        'submitting_user': str(metadata.get('user')),
        'output_path': str(metadata.get('output')),
        'cwd': str(metadata.get('cwd')),
        'hail_version': str(metadata.get('hailVersion')),
        'request_body': metadata.get('meta', {}) or {},
    }
```

dancoates (changes requested):
This looks good, there's just a couple of bits to update. Could you also please update the README in the metamist folder to correct the line about passing the `AUDIENCE_URL`? The correct env var is now `SM_URL` I'd generally refer to it as "metamist url" instead of "audience url"

dancoates (approved):
This looks good, just need to fix the lint errors and then all good to deploy

